PR #144, opened 2021-12-29
Files

ladnane:
- [ x] Bug fix
- [ x] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ x] No

* **Other information**:

### DIFF
--- a/dessia_common/core.py
+++ b/dessia_common/core.py
@@ -1509,10 +1509,7 @@ def deserialize_argument(type_, argument):
             msg = "Deserialization of typing {} is not implemented"
             raise NotImplementedError(msg.format(type_))
     elif type_ is TextIO:
-        try:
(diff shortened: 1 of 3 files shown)


ladnane:
* BytesIO tested : OK (return a BinaryFile with filename attribute, could be used for testing by PE ...)
* StringIO types : not working, we get the same typing from frontend as BytesIO  (dessia_common.typings.FileType)-> so we get a binary instead. we either need the correct type (maybe use dessia_common.typings.BinaryFileType + dessia_common.typings.TextFileType) to separate the two ?
* other types from dc.files => we need frontend to fetch file when (like forother files now) + see how to validate/ filter ... on backend.for nowu alwaysget either a BinaryFile or a StringFile

masfaraud (approved):

PR #172, opened 2022-01-24
Workflow export blocks

GhislainJ:
- [ ] Bug fix
- [x] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [x] No

* **Other information**:
This a first proposition of workflow exports as a zip file. See `workflow_export.py` script. Start Run -> Continue Run -> Export Archive.
Workflow state progress is 1 once output is reached, then >= 1 when exports are evaluated.
export_archive cannot be run if state progress is not >= 1

This uses the generic functions such as `save_to_file`, `to_step`, that write a file on disk and need a filename as inputs. For now, I gave a temp, hardcoded filepath as the inputs (which will probably break for complex archives), then add all the files into the archive, but it would be better to work with file-like objects. **Keeping on-disk file would require to return the filepath to created object for every method used, which is a breaking change**

I added types to blocks, but we could do differently (inheritance, ...). Export block need a method_type as its inputs.
An Archive block gather all the export block results (created file) into an archive. Export blocks cannot be used without an archive as for now

the export_archive yields a bytes-like object

Some code have been copy-pasted, so we might have some mututalization to perform

These have been coded into WorkflowState. To avoid code duplication into WorkflowRun, we should take a decision on how to manage them together. Do we remove WorkflowRun, do we make it inherit WorkflowState ?


dessia_common/workflow.py
```diff
@@ -2191,6 +2247,42 @@ def to_workflow_run(self, name=''):
         else:
             raise ValueError('Workflow not completed')
 
+    def _export_formats(self):
+        formats = []
+        for block in self.workflow._exportable_blocks:
```
masfaraud, line 2252: Count export types to have "zip_1" "zip_2", ... ?

dessia_common/workflow.py
```diff
@@ -2191,6 +2247,42 @@ def to_workflow_run(self, name=''):
...
+            if isinstance(block, Archive):
```
masfaraud, line 2253: make export block generate the export format

dessia_common/workflow.py
```diff
@@ -2191,6 +2247,42 @@ def to_workflow_run(self, name=''):
...
+                formats.append(('zip', 'export_archive', False))
+            # elif isinstance(block, Export):
+            #     formats.append(('', '', None))
+        return formats
+
+    def export_archive(self):
+        if self.progress >= 1:
+            archive_blocks = [isinstance(b, Archive) for b in self.workflow._exportable_blocks]
+            if not any(archive_blocks):
```
masfaraud, line 2262: Why this limit? why not evaluate all export blocks?

masfaraud (approved):

PR #173, opened 2022-01-28
To dict

masfaraud:
Fixing problem of to_dict that should not have some pointers


masfaraud:
embedding of code of another WIP PR

PR #175, opened 2022-02-02
fix: version number with commit hash, docs

masfaraud:
Setup enhancement


dessia_common/errors.py
```diff
@@ -11,7 +11,7 @@ def __init__(self, message, traceback_=''):
         self.traceback = traceback_
 
     def __str__(self):
-        return f'{self.message}\nTraceback:\n{self.traceback}'
+        return f'{self.args[0]}\nTraceback:\n{self.traceback}'
```
ladnane, line 14: I don't understand why this change.

PR #264, opened 2022-05-25
fix: foreachname configure not overriden anymore

xavier-ferry:
- [x] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [x] No

* **Other information**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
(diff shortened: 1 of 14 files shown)


GhislainJ (commented):
I am puzzled about these changes. I cannot seem to find what this is achieving and I actually think that this is faulty : in the case of ir == iter_input_index, `input_` would be undefined and we would try and append it to inputs

xavier-ferry:
Sure, it feels like a wrong indent. 
I fixed it

GhislainJ (approved):

PR #277, opened 2022-06-17
Feat/cluster

Tanguylo:
- [ ] Bug fix
- [x] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes: 
- [x] No

* **Other information**: 
* It is possible that this PR introduces a breaking change for clustering methods users but their code will work without this package
*


masfaraud (approved):

dessia_common/cluster.py
```diff
@@ -17,67 +17,209 @@ class ClusterResult(dc.DessiaObject):
     Cluster object to instantiate and compute clusters on data.
     """
     _standalone_in_db = True
+    _allowed_methods = ['from_agglomerative_clustering', 'from_kmeans', 'from_dbscan']
 
-    def __init__(self, data: List[dc.DessiaObject], labels: List[int], name: str = ''):
+    def __init__(self, data: List[dc.DessiaObject] or List[List] = None, labels: List[int] = None, name: str = ''):
```
masfaraud, line 22: not fan of list[list]

masfaraud (changes requested):

dessia_common/cluster.py
Tanguylo: Changed to List[List[float]]

masfaraud (approved):

PR #288, opened 2022-07-06
Feat/workflow to script

xavier-ferry:
- [ ] Bug fix
- [x] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [x] No

* **Other information**:


dessia_common/workflow/utils.py
```diff
@@ -0,0 +1,51 @@
...
+"""
+
+from typing import List, Dict
+
+
+class ToScriptElement:
+    """
+    Class meant to improve to_script readability
+    """
+    before_declaration: str = None
```
GhislainJ, line 16: I think there is a misunderstanding on the differences between dessia_common and js here.
In python, attributes defined here are class attributes and its not the same concept that JS's private attributes.

Not only, you don't need to define an instance attribute here under the class definition, it is also wrong if you do it to attach the attribute to the class and change its value depending on the instance.

For example, if a Car has a class attribute for the number of wheels, which is 4 for every cars, a Ferrari instance (let's say the F1) **cannot** overwrite its value based on the instance, which is what is done here

dessia_common/workflow/utils.py
```diff
@@ -0,0 +1,51 @@
...
+                script_imports += import_str
+
+        return script_imports
+
+    def get_import_dict(self) -> Dict[str, List[str]]:
+        imports_dict: Dict[str, List[str]] = {}
+        for c in self.imports:
+            module = '.'.join(c.split('.')[:-1])
+            class_ = c.split('.')[-1]
+            if imports_dict.get(module) is None:
```
GhislainJ, line 45: It might be more pythonic and clear to do it this way : 

`if module in imports_dict:`

tests/workflow/workflow_with_models.py
```diff
@@ -76,3 +78,41 @@
 dict_from_json = json.loads(demo_workflow_json)
 deserialized_demo_workflow = wf.Workflow.dict_to_object(dict_from_json)
 assert demo_workflow == deserialized_demo_workflow
+
```
GhislainJ, line 81: Maybe we should split this and set the test of to_script in a separate test file.
You can still import a workflow from dessia_common/models/workflows

tests/workflow/workflow_with_models.py
```diff
@@ -76,3 +78,41 @@
...
+
+# Testing to_script
+demo_workflow.to_script()
```
GhislainJ, line 84: What would be really nice is to test the expected value with the `assert` keyword to check if we get what we expect, instead of only testing if the function is truthy

dessia_common/workflow/blocks.py
```diff
@@ -112,6 +113,18 @@ def dict_to_object(cls, dict_: JsonSerializable, force_generic: bool = False,
     def evaluate(_):
         return []
 
+    def _to_script(self) -> ToScriptElement:
+        input_variables = [f"Variable(memorize={input.memorize}, name='{input.name}')" for input in self.inputs]
+        inputs = '[' + ",".join(input_variables) + ']'
```
GhislainJ, line 118: Adding a space after the coma would match PEP 8 convention for the resulting script

dessia_common/workflow/blocks.py
```diff
@@ -112,6 +113,18 @@ def dict_to_object(cls, dict_: JsonSerializable, force_generic: bool = False,
...
+        script = f"Display(inputs={inputs}, order={self.order}, name='{self.name}')"
```
GhislainJ, line 119: I am unsure we wish to set inputs this way. A None value for inputs would mean that we are trying to add all the displays of an object. A defined value for inputs would mean it is used as a Base Class and we are for example instantiating a Multiplot Block.

This means we might need to split these two behaviors. I will have a look at it when working on Exports and Displays, soon

dessia_common/workflow/blocks.py
```diff
@@ -417,6 +458,18 @@ def _docstring(self):
                     docstring[input_] = block_docstring[input_]
         return docstring
 
+    def _to_script(self) -> ToScriptElement:
+        PREFIX = 'sub_'
```
GhislainJ, line 462: If this is a constant, it should be placed at module level, and in this case, not passed as an argument.

But I guess what you want to achieve is to customize its value depending on the subworkflow we are working on. What happens if there are several subworkflows ?

GhislainJ (changes requested):

dessia_common/workflow/blocks.py
GhislainJ: Alright, in this case, I guess, `prefix` is sufficient. it feels weird pythonically speaking to have a constant defined in a method, even if it means locally in the method scope. But I am a not technically against it, it just triggered something weird that got me worried. Like setting the name all caps would make the value instantly available at module level.

But as your exemple shows, it works perfectly

xavier-ferry: I set it back to lowercase `prefix`

GhislainJ (approved):

PR #315, opened 2022-08-01
fix(workflow): prevent non serialisable attributes in values and prop…

GhislainJ:
…erly deserialize workflow in wfs

- [x] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:

### DIFF
--- a/dessia_common/workflow/core.py
+++ b/dessia_common/workflow/core.py
@@ -21,7 +21,7 @@ from dessia_common import DessiaObject, is_sequence, JSONSCHEMA_HEADER, jsonsche
     deserialize_argument, set_default_value, prettyname, serialize_dict, DisplaySetting
 
 from dessia_common.utils.serialization import deserialize, serialize_with_pointers, serialize, update_pointers_data
-from dessia_common.utils.types import serialize_typing, deserialize_typing, recursive_type, typematch
+from dessia_common.utils.types import serialize_typing, deserialize_typing, recursive_type, typematch, is_serializable
 from dessia_common.utils.copy import deepcopy_value
 from dessia_common.utils.docstrings import FAILED_ATTRIBUTE_PARSING, EMPTY_PARSED_ATTRIBUTE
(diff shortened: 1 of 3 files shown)


dessia_common/workflow/core.py
```diff
@@ -1470,7 +1470,8 @@ def dict_to_object(cls, dict_: JsonSerializable, force_generic: bool = False,
         global_dict, pointers_memo = update_pointers_data(global_dict=global_dict, current_dict=dict_,
                                                           pointers_memo=pointers_memo)
 
-        workflow = Workflow.dict_to_object(dict_['workflow'])
+        workflow = Workflow.dict_to_object(dict_=dict_['workflow'], global_dict=global_dict,
```
xavier-ferry, line 1473: Sadly, I don't have the knowledge to quickly have a look at this modification

dessia_common/workflow/core.py
```diff
@@ -1799,9 +1800,13 @@ def __init__(self, workflow: Workflow, input_values, output_value, values,
             end_time = time.time()
         self.end_time = end_time
         self.execution_time = end_time - start_time
-        self.variable_values = {workflow.variable_indices(k): v for k, v in values.items() if k.memorize}
-        WorkflowState.__init__(self, workflow=workflow, input_values=input_values, activated_items=activated_items,
-                               values=values, start_time=start_time, output_value=output_value, log=log, name=name)
+        self.variable_values = {workflow.variable_indices(k): v for k, v in values.items()
```
xavier-ferry, line 1803: Is it enough to just filter ? 
I mean, do we want to do something about non-serializable values ?

dessia_common/workflow/core.py
GhislainJ: It's actually a wonder how workflow deserialization could work without these changes. I guess it didn't in some case, and did in some other, but it should never have worked without these new args

PR #328, opened 2022-08-10
Tests/optimiser

Tanguylo:
- [x] performance 

- [x] No

* **Other information**:
Implement tests for optimizer.
Some elements in code optimizer are not tested here because they do not seem to be used. Should they be removed ?


dessia_common/optimization.py
```diff
@@ -124,33 +121,41 @@ def cma_bounds(self):
...
 
         model = self.instantiate_model(attributes_values)
-
         return model, fx_opt
+
+    def optimize_cma_then_gradient(self, method: str = 'L-BFGS-B'):
+        model_cma, fx_cma = self.optimize_cma()
+        x0 = npy.array([getattr(model_cma, attr.attribute_name) for attr in self.optimization_bounds])
+        model, fx_opt = self.optimize_gradient(method=method, x0=x0)
+        best_fx = npy.min([fx_opt, fx_cma])
```
masfaraud, line 157: why not just if fx_opt < fx_cma:

dessia_common/optimization.py
```diff
@@ -124,33 +121,41 @@ def cma_bounds(self):
...
+        result = scipy.optimize.minimize(self.objective_from_dimensionless_vector, x0, bounds=bounds, method=method)
 
+        attributes_values = self.vector_to_attributes_values(self.dimensionless_vector_to_vector(result.x))
         model = self.instantiate_model(attributes_values)
         return model, result.fun
 
-    def optimize_cma(self):
+    def _init_gradient(self, x0: List[float] = None):
+        if x0 is None:
+            return npy.random.random(self.number_parameters)
```
masfaraud, line 136: a little bit overkill. why not a generic method random_x ?

masfaraud (changes requested):

dessia_common/optimization.py
Tanguylo: Modification made:

> model, best_fx = self.optimize_gradient(method=method, x0=x0)
>  if fx_cma <= best_fx:
>       model = model_cma
>       best_fx = fx_cma
> return model, best_fx

dessia_common/optimization.py
Tanguylo: Merged method _init_gradient into optimize_gradient

masfaraud (approved):

PR #339, opened 2022-08-31
fix: received typing error when its not a class

ladnane:
- [x] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [x
image (4)
] No

* **Other information**:


GhislainJ:
I am afraid that this might not be as simple as only checking if typing is a class. I went across this bug several times and it was almost everytime a wrong typing usecase. For example something like 'Any' reaching this. I that extent checking for is_class might unbalance elifs structure and bring othe problems later.

Another example, once it was just a problem of "if orders". I would prefer to have an overview of what is causing the issue in order to merge this

ladnane:
the 'Any' case is checked before, so this can't be the cause of the pb. 
my uderstanding of the code is that the typing can be a string / obj in wich case we handle this in the last 'else':
```   
   elif issubclass(typing_, (BinaryFile, StringFile)):
        jsonschema_element[key].update({'type': 'text', 'is_file': True})
    else:
        classname = dc.full_classname(object_=typing_, compute_for='class')
        if inspect.isclass(typing_) and issubclass(typing_, dc.DessiaObject):
            # Dessia custom classes
            jsonschema_element[key].update({
                'type': 'object',
                'standalone_in_db': typing_._standalone_in_db
            })
        else:
            # Statically created dict structure
            jsonschema_element[key].update(static_dict_jsonschema(typing_))
        jsonschema_element[key]['classes'] = [classname]
    return jsonschema_element
```
without this change the last 'else' is never executed. also, we will at least get a better error message imo.

PR #351, opened 2022-09-12
some debugs on serialization & memos

masfaraud:
Debugging alias of class


code_pylint.py
```diff
@@ -27,6 +27,9 @@
                      'too-many-locals': 12,  # Reduce by dropping vectored objects
                      'wrong-import-order': 7,
                      'too-many-branches': 10,
+                     'unused-import': 1,
```
GhislainJ, line 30: duplicate keys in dict

GhislainJ (changes requested):

PR #364, opened 2022-09-20
Feat/plot data block

GhislainJ:
- Add block for plot_data
- Avoid empty values when display/export blocks are defined
- Try and mutualize simple display blocks code
- Add some test about all of the above

* **Other information**:


Launayl (approved):
That resolve the multiplot bug In my case

PR #374, opened 2022-09-26
Feat/sampling

Tanguylo:
- [ ] Bug fix
- [x] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [x] No

* **Other information**:
Implements a Sampler object so that it is easy to build a DOE with sampling methods


Tanguylo:
Should be reviewed after merging fix/hlist_* PR for better reading

dessia_common/optimization.py
```diff
@@ -26,11 +26,12 @@ def __init__(self, attribute_name: str, value: float, name: str = ''):
 class BoundedAttributeValue(dc.DessiaObject):
     _standalone_in_db = True
 
-    def __init__(self, attribute_name: str, min_value: float, max_value: float, name: str = ''):
+    def __init__(self, attribute_name: str, min_value: float, max_value: float, number: int = 2, name: str = ''):
```
masfaraud, line 29: Why this number?

dessia_common/sampling.py
```diff
@@ -0,0 +1,144 @@
...
+        varying_sampling = pyDOE.lhs(len(self.sampled_attributes), samples=samples, criterion=criterion)
+        full_doe = []
+        fixed_values = [attr.value for attr in self.constant_attributes]
+        for nodim_sample in varying_sampling:
+            varying_values = [attr.dimensionless_to_value(value)
+                              for attr, value in zip(self.sampled_attributes, nodim_sample)]
+            full_doe.append(self.sampled_class(**dict(zip(self.attributes, fixed_values + varying_values))))
+        return full_doe
+
+    def _montecarlo_sampling(self, samples: int = 10):
```
masfaraud, line 79: remove default value

dessia_common/sampling.py
```diff
@@ -0,0 +1,144 @@
...
+        instances_numbers = self._get_instances_numbers()
+        parameter_grid = self._build_parameter_grid(instances_numbers)
+        idx_sampling = pyDOE.fullfact(instances_numbers)
+        full_doe = []
+        for idx_sample in idx_sampling:
+            valued_sample = [parameter_grid[attr_row][int(idx)] for attr_row, idx in enumerate(idx_sample)]
+            full_doe.append(self.sampled_class(**dict(zip(self.attributes, valued_sample))))
+        return full_doe
+
+    def _lhs_sampling(self, samples: int = 10, criterion: str = 'center'):
```
masfaraud, line 69: remove default value for sample

dessia_common/sampling.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def _montecarlo_sampling(self, samples: int = 10):
+        full_doe = []
+        fixed_values = [attr.value for attr in self.constant_attributes]
+        for _ in range(samples):
+            varying_values = [random.uniform(attr.min_value, attr.max_value) for attr in self.sampled_attributes]
+            full_doe.append(self.sampled_class(**dict(zip(self.attributes, fixed_values + varying_values))))
+        return full_doe
+
+    def _get_doe(self, method: str = 'fullfact', samples: int = None, lhs_criterion: str = 'center'):
```
masfaraud, line 87: Why int default is None?

masfaraud (changes requested):

dessia_common/sampling.py
Tanguylo: done

dessia_common/sampling.py
Tanguylo: done

dessia_common/sampling.py
Tanguylo: Put at 10, it was a mistake

dessia_common/optimization.py
Tanguylo: It has been added to specify the number of values the user wants to generate for this BoundedAttributeValue.
It is used here:
```
def _get_instances_numbers(self):
       return [1] * len(self.constant_attributes) + [attr.number for attr in self.sampled_attributes]
```

The two other options I see is: 

1. to specify directly in the class Sampler with a list of float, of the same length of attribute 'sampled_attributes'
2. to create a new class that inherits from BoundedAttributeValue and is only used in Sampler, with this attribute 'number'

dessia_common/sampling.py
```diff
@@ -93,7 +93,7 @@ def _get_doe(self, method: str = 'fullfact', samples: int = None, lhs_criterion:
             return self._montecarlo_sampling(samples=samples)
         raise NotImplementedError(f"Method '{method}' is not implemented in {self.__class__}._get_doe method.")
 
-    def make_doe(self, method: str = 'fullfact', samples: int = None, lhs_criterion: str = 'center', name: str = ''):
+    def make_doe(self, method: str = 'fullfact', samples: int = 10, lhs_criterion: str = 'center', name: str = ''):
```
masfaraud, line 96: Delete default 10 values

masfaraud (changes requested):

dessia_common/sampling.py
Tanguylo: Removed and fixed one mistake.
Do we keep defaults value for method ?

masfaraud (approved):

PR #382, opened 2022-10-04
wip: debug in progress

masfaraud:
Left behind branch: opening a PR to decide if this code is worth merging
Do a real review


dessia_common/utils/diff.py
```diff
@@ -38,11 +38,19 @@ def diff(value1, value2, path='#'):
...
-        if value1._data_eq(value2):
+    else:
+        # Should be object
+        # print(value1, isinstance(value1, dc.DessiaObject),  hasattr(value1, '_data_eq'))
+        # if isinstance(value1, dc.DessiaObject):
+        if hasattr(value1, '_data_eq'):
+            # DessiaObject
+            if value1._data_eq(value2):
+                return [], [], []
+        elif hasattr(value2, '_data_eq'):
```
GhislainJ, line 49: This case looks weird. If value 2 have _data_eq that value1 do not, hence they are necessarily different, right ? It looks like we should never end up in here

masfaraud:
Too old now, some bug fix have been done

PR #392, opened 2022-10-07
Fix/layout

xavier-ferry:
- [x] new features

- [ ] Yes:
- [ ] No

* **Other information**:


dessia_common/graph.py
```diff
@@ -56,3 +56,43 @@ def extract_region(networkx_graph: nx.Graph, nodes, distance: int = 5):
         region_nodes.extend(ancs)
 
     return networkx_graph.subgraph(region_nodes)
+
+
+def get_longest_path(graph: nx.DiGraph, begin, end) -> list:
+    return list(nx.shortest_simple_paths(graph, begin, end))[-1]
+
+
+def get_paths_from_to(graph: nx.DiGraph, _from: list, _to: list) -> list:
```
GhislainJ, line 65: We have to be careful with types here. `list` is ambiguous over python version and should not be used (use `List`, instead). In addition to that, it should never be used without an argument (`List[SomeType]`), otherwise it is subject to errors. 

In any case, this is not mandatory as no jsonschema would be created from this, so I even prefer no typing (and thus no weird and unwanted behavior) over incomplete ones.

This goes for other parts of the code (get longest_path,...) and Dicts as well (get_distance_by_nodes,...)

dessia_common/graph.py
```diff
@@ -56,3 +56,43 @@ def extract_region(networkx_graph: nx.Graph, nodes, distance: int = 5):
...
+
+    for node in longest_path:
+        distances[node] = longest_path.index(node)
+
+    for node in [n for n in graph.nodes if n not in longest_path]:
+        try:
+            path = get_longest_path(graph, node, end_of_path)
+            distances[node] = len(longest_path) - len(path)
+        except nx.exception.NetworkXNoPath:
+            pass
```
GhislainJ, line 90: What means this exception and why are we avoiding it ?

dessia_common/graph.py
```diff
@@ -56,3 +56,43 @@ def extract_region(networkx_graph: nx.Graph, nodes, distance: int = 5):
...
+    for origin in _from:
+        for destination in _to:
+            try:
+                paths.append(get_longest_path(graph, origin, destination))
+            except nx.exception.NetworkXNoPath:
+                continue
+    return paths
+
+
+def get_distance_by_nodes(graph: nx.DiGraph) -> Dict:
```
GhislainJ, line 76: I think some comments (and/or setting some variables with explicit names, for ex, for list comprehensions) may greatly improve the readibility and the understanding of this bit of code here

dessia_common/workflow/blocks.py
```diff
@@ -187,7 +187,7 @@ def dict_to_object(cls, dict_: JsonSerializable, force_generic: bool = False,
         class_ = get_python_class_from_class_name(classname)
         name = dict_['name']
         method_type = ClassMethodType(class_=class_, name=method_name)
-        return cls(method_type=method_type, name=name)
+        return cls(method_type=method_type, name=name, position=dict_.get('position'))
```
GhislainJ, line 190: Why not just `position=dict_["position"]` ? Seems more readable to me, as we do not define any default value to the get.

This goes for all the others, as well

dessia_common/workflow/core.py
```diff
@@ -163,13 +183,16 @@ class Block(DessiaObject):
     _non_serializable_attributes = []
 
     def __init__(self, inputs: List[Variable], outputs: List[Variable],
-                 position: Tuple[float, float] = (0, 0), name: str = ''):
+                 position: Tuple[float, float] = None, name: str = ''):
```
GhislainJ, line 186: Tuple default assignment works perfectly fine as these are not mutables

dessia_common/workflow/core.py
```diff
@@ -1101,53 +1131,43 @@ def variable_compatibility(self, variable: Variable, other_variable: Variable) -
...
+            if pipe.input_variable in self.nonblock_variables:
+                input_node = pipe.input_variable
+            else:
+                input_node = self.block_from_variable(pipe.input_variable)
+            output_block = self.block_from_variable(pipe.output_variable)
+            graph.add_edge(input_node, output_block)
+
+        return graph
+
+    def graph_distances(self, graph) -> list:
```
GhislainJ, line 1149: Same remark as before, here

GhislainJ (changes requested):

dessia_common/graph.py
xavier-ferry: This exception is raised when there is no path between `node` and `end_of_path`. 
In such a case, we look for a path from any treated node to `node`

In fact, we could write this, but I'm not sure that this is more readable: 
```py 
    for node in [n for n in graph.nodes if n not in longest_path]:
        try:
            path = get_longest_path(graph, node, end_of_path)
            distances[node] = len(longest_path) - len(path)
        except nx.exception.NetworkXNoPath:
            for path in get_paths_from_to(graph, list(graph.nodes), [node]):
                distance_via_current_node = distances[path[0]] + len(path) - 1
                if node not in distances or distances[node] < distance_via_current_node:
                    distances[node] = distance_via_current_node
```

dessia_common/workflow/blocks.py
xavier-ferry: `dict_['position']` raises an error if position is not in dict_
Whereas dict_.get('position') returns None in such a case.

The point of this dev was to call the init with None if no explicit position has been set. 
In this case the init changes the None value for (0,0)
-> This answer another of your comment where you say that I can directly write 
```py
    def __init__(self, inputs: List[Variable], outputs: List[Variable],
                 position: Tuple[float, float] = (0,0), name: str = ''):
```

dessia_common/workflow/core.py
xavier-ferry: See the answer just above

dessia_common/workflow/blocks.py
GhislainJ: Alright, I thought one was supposed to write `dict_.get("stuff", default_value)` to have a default value, even for None.
That seems all good to me, then !

GhislainJ (approved):

PR #393, opened 2022-10-10
Workflow & Blocks general fixes

GhislainJ:
- [x] Bug fix
- [x] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [x] No

* **Other information**:
- Fixes CadView block type
- Add filename input to export blocks
- Add ConcatString block
- Handle streams with filename + extensions

### DIFF
--- a/dessia_common/workflow/blocks.py
+++ b/dessia_common/workflow/blocks.py
@@ -978,6 +978,53 @@ class Substraction(Block):
         return ToScriptElement(declaration=script, imports=[self.full_classname])
 
 
+class ConcatenateStrings(Block):
+    """
+    Concatenates the n input elements, separate by the separator input, into one string
+
+    :param number_elements: Number of block inputs
+    :type number_elements: int
+    :param separator: Character used to joins the input elements together
+    :type separator: str
+    """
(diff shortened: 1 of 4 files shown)


dessia_common/workflow/blocks.py
```diff
@@ -963,6 +978,53 @@ def _to_script(self) -> ToScriptElement:
         return ToScriptElement(declaration=script, imports=[self.full_classname])
 
 
+class ConcatenateStrings(Block):
```
xavier-ferry, line 981: In https://github.com/Dessia-tech/dessia_common/pull/392, position has been added to __init__ and to_script has been improved. 

If this PR is merged before the one just above, we will need to update 392 to handle position in init and in to_script. 
If not, we should update this PR one 392 is merged

xavier-ferry (commented):
Tbh, I did not looked in detail the display_settings / selector / type_ part as I don't have any particular knowledge about this.

dessia_common/workflow/blocks.py
GhislainJ: This is to be merged directly into master, so it will happen on a eventual chore/testing-to-dev branch

PR #397, opened 2022-10-12
Feat/generic export

masfaraud:
- [ ] Bug fix
- [x] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [x] No

Implement a generic save to file from selector
Also creating a selector

* **Other information**:


dessia_common/core.py
```diff
@@ -638,10 +638,28 @@ def to_xlsx_stream(self, stream):
         writer.save_to_stream(stream)
 
     def _export_formats(self):
-        formats = [{"extension": "json", "method_name": "save_to_stream", "text": True, "args": {}},
-                   {"extension": "xlsx", "method_name": "to_xlsx_stream", "text": False, "args": {}}]
+        formats = [{"selector": "json", "extension": "json",
```
GhislainJ, line 641: Workflows now define an export_format with an "export_name" entry (that could be renamed "filename"), in order to set a by default filename known by frontend for an improved experience while exporting stuff.

Should we set it in this generic definition ? If yes, which names could we compute by default ?

dessia_common/core.py
```diff
@@ -638,10 +638,28 @@ def to_xlsx_stream(self, stream):
...
+                   {"selector": "xlsx", "extension": "xlsx",
+                    "method_name": "to_xlsx_stream", "text": False, "args": {}}]
         return formats
 
+    def save_export_to_file(self, selector, filepath):
+        for export_format in self._export_formats():
+            if export_format['selector'] == selector:
+                if not filepath.endswith(f".{export_format['extension']}"):
+                    filepath += f".{export_format['extension']}"
+                    print(f'Renaming filepath to {filepath}')
```
Tanguylo, line 652: Here is a suggestion, not sure it is better
You can replace it with : 
```
extension = f".{export_format['extension']}"
filepath += f".{extension}" * (not filepath.endswith(extension))
```

dessia_common/core.py
```diff
@@ -638,10 +638,28 @@ def to_xlsx_stream(self, stream):
...
+                if export_format['text']:
+                    mode = 'w'
+                else:
+                    mode = 'wb'
```
Tanguylo, line 656: Here is a suggestion, not sure it is better
you can do it :
`mode = 'wb'[:1 + bool(export_format['text'])]`

dessia_common/core.py
GhislainJ: I do really find it hard to read. The later is really straightforward and almost English, as python should be written. It literally says "If filepath does not end with extension, well, we add it".

Your solution can be sophisticated in a way, but one must just stop one or two sec to properly understand it, which is a big flaw, in my opinion

dessia_common/core.py
GhislainJ: Same remark as the last comment, here. I find it really complex for simple stuff.

I personnally think that the right solution here would be : 
```
mode = "w"                              # or whatever you want it to be
if not export_format["text"]:
    mode += "b"                         # or mode = f"{mode}b"
```

GhislainJ:
I am currently adding changes to this, to support workflows as well

dessia_common/exports.py
```diff
@@ -41,6 +41,23 @@ def is_builtins_list(list_):
     return True
 
 
+class ExportFormat:
+    def __init__(self, selector: Optional[str], extension: str, method_name: str, text: bool,
+                 export_name: str = "", args: Dict[str, Any] = None):
+        self.selector = selector
+        self.extension = extension
+        self.method_name = method_name
+        self.text = text
```
Tanguylo, line 50: It is not so obvious that text is a boolean. Maybe "with_text" could be better...

dessia_common/forms.py
```diff
@@ -48,23 +48,43 @@
...
+        self.intarg = intarg
+        self.boolarg = boolarg
 
         PhysicalObject.__init__(self, name=name)
 
     @classmethod
-    def generate(cls, seed: int) -> 'StandaloneSubobject':
-        floatarg = Distance(1.7 * seed)
+    def generate(cls, seed: int) -> 'EmbeddedBuiltinsSubobject':
+        floatarg = 0.3
```
Tanguylo, line 79: You could generate a random number here.
This may avoid some future bugs because a check for a specific float is not as good as with a generic one

dessia_common/forms.py
```diff
@@ -48,23 +48,43 @@
...
+        distarg = Distance(1.7 * floatarg * seed)
+        intarg = seed
```
Tanguylo, line 81: Idem, why not make a randint and get rid of seed argument ?

dessia_common/forms.py
```diff
@@ -88,40 +108,66 @@ def voldmlr_primitives(self):
...
-        StandaloneSubobject.__init__(self, floatarg=floatarg, name=name)
+
+class EnhancedStandaloneSubobject(StandaloneBuiltinsSubobject):
+    def __init__(self, floatarg: Distance, name: str = 'Standalone Subobject'):
+        StandaloneBuiltinsSubobject.__init__(self, distarg=floatarg, floatarg=floatarg, intarg=floor(floatarg),
+                                             boolarg=floatarg.is_integer(), name=name)
 
     @classmethod
     def generate(cls, seed: int) -> 'EnhancedStandaloneSubobject':
         floatarg = Distance(1.2 * seed)
```
Tanguylo, line 146: random seed ?

dessia_common/forms.py
```diff
@@ -88,40 +108,66 @@ def voldmlr_primitives(self):
...
-        StandaloneSubobject.__init__(self, floatarg=floatarg, name=name)
+        StandaloneBuiltinsSubobject.__init__(self, distarg=distarg, floatarg=floatarg, intarg=intarg,
+                                             boolarg=boolarg, name=name)
 
     @classmethod
     def generate(cls, seed: int) -> 'InheritingStandaloneSubobject':
-        floatarg = Distance(0.7 * seed)
+        distarg = Distance(0.7 * seed)
+        floatarg = 0.1 * seed
         strarg = str(-seed)
```
Tanguylo, line 166: str(-seed) ? Why not str(seed) ?

dessia_common/workflow/core.py
```diff
@@ -389,18 +390,18 @@ def __init__(self, blocks, pipes, output, *, imposed_variable_values=None,
 
         self.block_selectors = {}
         for i, block in enumerate(self.blocks):
```
Tanguylo, line 392: idx instead of i

dessia_common/workflow/core.py
```diff
@@ -543,6 +544,17 @@ def copy_pipes(self, copied_workflow: 'Workflow') -> List[Pipe]:
...
 
+    def branch_by_selector(self, blocks: List[Block]):
+        """
+        Return the corresponding branch to each display or export selector
+        """
+        selector_branches = {}
+        for block in blocks:
+            branch = self.secondary_branch_blocks(block)
+            selector = self.block_selectors[block]
+            selector_branches[selector] = branch
```
Tanguylo, line 555: why not just : 
`selector_branches[self.block_selectors[block]] = self.secondary_branch_blocks(block)
`

tests/jsonschema.py
```diff
@@ -219,8 +205,9 @@
 
 subobject_default_value = jss.chose_default(jsonschema["properties"]["standalone_subobject"])
 assert subobject_default_value["name"] == "StandaloneSubobject1"
-assert subobject_default_value["object_class"] == "dessia_common.forms.StandaloneSubobject"
-assert subobject_default_value["floatarg"]['value'] == 1.7
+assert subobject_default_value["object_class"] == "dessia_common.forms.StandaloneBuiltinsSubobject"
+assert subobject_default_value["floatarg"] == 0.3
+assert subobject_default_value["distarg"]["value"] == 0.51
```
Tanguylo, line 210: with rand values, these tests won't work anymore

tests/jsonschema.py
```diff
@@ -242,7 +227,8 @@
 assert subobject_default_value["name"] == "Inheriting Standalone Subobject1"
 assert subobject_default_value["object_class"] == "dessia_common.forms.InheritingStandaloneSubobject"
 assert subobject_default_value["strarg"] == "-1"
-assert subobject_default_value["floatarg"]['value'] == 0.7
+assert subobject_default_value["floatarg"] == 0.1
+assert subobject_default_value["distarg"]["value"] == 0.7
```
Tanguylo, line 231: idem if rand values are used

dessia_common/exports.py
GhislainJ: Yes, I do agree. In my opinion, the best solution would be is_text, is_binary, is_text_like, is_binary_like.

Unfortunately, this has a lot of implications on other repos and would be a breaking change. So we will probably keep it until we have a major version release

dessia_common/forms.py
GhislainJ: I think it would be difficult to test it afterwards (we check its value in dessia_common/tests/unit_tests and/or some other test files)

dessia_common/forms.py
GhislainJ: Here, it serves the only purpose on testing frontend and platform behavior.

dessia_common/workflow/core.py
GhislainJ: I cannot find the docs on it again, but as long as the convention didn't change, i is allowed by it and its use is even encouraged. Just like we should use x as a lambda variable or one-letter var in comprehension list. 

If you find some resources stating otherwise (because I cannot seem to find anything), I would gladly have a look at it, but I don't think we will initiate these kind of changes because we use the "i" absolutely everywhere

dessia_common/forms.py
GhislainJ: The seed arg does have a lot of pros, actually. At some point, I am building a boolean over the fact that seed is even or not. So it gives us ways to check some stuff for example in serialization/deserialization, depending on a certain value, and this even on the platform.

dessia_common/exports.py
GhislainJ: In fact, I also wanted to change sur "export_name" stuff to "filename" but didn't do it for the same reasons.
The sole purpose of not doing it now is because adding breaking changes by refactoring and adding new feature is a quite dangerous operation. And I already refactored the ExportFormat from a dict to an object while adding selector to it. So I was not willing to take more risks of breaking everything

dessia_common/workflow/core.py
GhislainJ: I'll call personal preference on this one. I wouldn't mind if I saw what you propose in a code review (or at least, i wouldn't mind if i didn't think splitting on several lines would greatly increase readability), but i find it easier to read and understand when splitting it as a three-liner while naming the concepts. As the function is not that big, it is not a big deal to have it like this imo

dessia_common/workflow/core.py
Tanguylo: Actually, pylint was set not to be happy when the iterator is only one letter long.
I guess this option has since been disabled :)

PR #398, opened 2022-10-14
Feat/markdowns2

Tanguylo:
- [ ] Bug fix
- [x] new features
- [ ] performance 
- [ ] docs update

* **Other information**:
Make a better default markdown for DessiaObjects and Datasets.
Associated with "feat/md_table_style" branch of frontend repo.

image

image


Tanguylo:
New object MarkdownWriter able to : 

- Create a matrix table from a matrix and titles
- Create a table from an object
- Standardizes 'to string' for markdowns

This PR is associated to the [frontend PR](https://gitea.dessia.tech/DessiA/frontend/pulls/1332 ) that modifies the table styles for markdowns : 
Maybe some work on fonts and alignment is still required

image
image

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+
+    def _head_table(self, col_names: List[str] = None) -> str:
+        cap_names = map(lambda x: x.capitalize(), col_names)
+        return ("| " + " | ".join(cap_names) + " |\n" +
+                "| ------ " * len(col_names) + "|\n")
+
+    def _table_line(self, row: List[Any]) -> str:
+        line = "|"
+        for value in row:
+            line += ''.join([" ", self._string_in_table(self._value_to_str(value)), " |"])
```
GhislainJ, line 322: Why not `" | ".join([self._string_in_table(self._value_to_str(value))])` ?

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+        return line + "\n"
+
+    def _content_table(self, content: List[List[Any]]) -> str:
+        half_table = int(self.table_limit / 2)
+        table = ''
+        for row in content[:half_table]:
+            table += self._table_line(row)
+
+        if len(content) > self.table_limit and self.table_limit != -1:
```
GhislainJ, line 331: -1 stands for no limit ? Why not None in this case ?

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+            for row in content[-half_table:]:
+                table += self._table_line(row)
+
+        return table
+
+    def print_name(self, object_) -> str:
+        return (object_.name if object_.name != '' else 'with no name')
+
+    def print_class(self, object_) -> str:
+        return object_.__class__.__name__
```
GhislainJ, line 344: Could make use of the previously commented `classname` function

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+
+    def matrix_table(self, matrix: List[List[float]], col_names: List[str] = None) -> str:
+        return ''.join([self._head_table(col_names),
```
GhislainJ, line 347: Does this line really need to be cut in half ? ie. doesn't it fit in the 120 characters limit  if kept as a one-liner ?

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+                        self._content_table(matrix)])
+
+    def object_table(self, object_) -> str:
+        return self.matrix_table(self._object_matrix(object_),
```
GhislainJ, line 351: Same here

dessia_common/forms.py
```diff
@@ -378,80 +378,80 @@ def maldefined_method(self, arg0, arg1=1, arg2: int = 10, arg3=3):
     # def method_with_faulty_typing(arg0: Iterator[int]):
     #     return arg0
 
-    def to_markdown(self):
```
GhislainJ, line 381: At first, I was OK to remove all this, but then I realised that it was a cool way to show how titles, lists, links, etc behave on the platform.

Couldn't we keep it and concatenate the super method to add the table at the bottom of it ?

dessia_common/templates.py
```diff
@@ -322,14 +327,18 @@
 ''')
 
 
-Dataset_markdown_template = Template('''
-# Object $name of class $class_
+dataset_markdown_template = Template('''
```
GhislainJ, line 330: THANK YOU ! What a horrible mess of a variable name it was

tests/markdowns.py
```diff
@@ -0,0 +1,20 @@
+"""
```
GhislainJ, line 1: Do we actually make tests here ? We could simply check the str length to ensure there are no regressions when the code will be changed in the future.

Right now, it looks like it serves the only purpose to increase coverage

GhislainJ (changes requested):

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+
+class MarkdownWriter:
+    def __init__(self, print_limit: int = 25, table_limit: int = 12):
+        self.print_limit = print_limit
+        self.table_limit = table_limit
+
+    def _object_titles(self):
+        return ['Attribute', 'Type', 'Value'] #, 'Subvalues']
+
+    def _sequence_to_str(self, value: List[Union[list, dict, set]]):
```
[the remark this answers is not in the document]
Tanguylo: Thanks, it's done

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+        in_values = value
+        if len(value) == 0:
+            return f"empty {type(value).__name__}"
+
+        printed_string = f"{len(value)} "
+        if isinstance(value, dict):
```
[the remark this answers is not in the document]
Tanguylo: It is to take its values in the next row
```
printed_string = f"{len(value)} "
if isinstance(value, dict):
    in_values = list(value.values())
```

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+            in_values = list(value.values())
```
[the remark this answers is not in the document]
Tanguylo: No it is not really a problem because it is just a print.
The purpose here is to get the string of all classes in printed_string (e.g. : `printed_string = 'class1, class2, class3') and to print the first self.print_limit char of this string in the table

dessia_common/exports.py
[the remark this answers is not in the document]
GhislainJ: Yes, but the fact that value is typed as a List in the annotations is inconsistent with the fact that we treat it as a Dict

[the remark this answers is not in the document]
Tanguylo: image

I changed to ` def _sequence_to_str(self, value: Union[List, Dict, Set]):`

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+            printed_string += ''.join([f"{all_class_names[0]}",
+                                       f"{'s' if len(value) > 1 else ''}"])
+        else:
+            str_all_class = str(all_class_names).translate(str(all_class_names).maketrans('', '', "{}'"))
+            printed_string += f"{str_all_class}"
+
+        return printed_string
+
+    def _object_to_str(self, value) -> str:
+        if hasattr(value, 'name' ):
```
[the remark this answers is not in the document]
Tanguylo: YESSSS !

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+            return f"{value.name}"
+        return 'unnamed'
+
+    def _value_to_str(self, value: Any) -> str:
+        if isinstance(value, (float, int, bool, complex)):
+            return str(round(value, 6))
+        if isinstance(value, str):
+            return (value if value != '' else 'no_value')
+        if isinstance(value, (list, dict, set)):
```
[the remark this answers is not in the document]
Tanguylo: I replaced with is_sequence

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+            return self._sequence_to_str(value)
+        return self._object_to_str(value)
```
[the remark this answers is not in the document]
Tanguylo: like it ? :

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+
+    def _string_in_table(self, string: str = ''):
+        return string[:self.print_limit] + ('...' if len(string) > self.print_limit else '')
+
+    def _object_matrix(self, object_):
+        matrix = []
+        for attr, value in object_.__dict__.items():
+            matrix.append([attr,
+                           type(value).__name__,
```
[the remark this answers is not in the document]
Tanguylo: We spoke about it, i made the modif for type

dessia_common/exports.py
```diff
@@ -253,3 +256,100 @@ def autosize_sheet_columns(sheet, min_width=5, max_width=30):
...
+                           # (self._sequence_to_str(value) if isinstance(value, (list, dict, set)) else ' - '),
+                           (self._value_to_str(value) if not isinstance(value, (list, dict, set)) else ' - ')])
+        return matrix
+
+    def _head_table(self, col_names: List[str] = None) -> str:
+        cap_names = map(lambda x: x.capitalize(), col_names)
```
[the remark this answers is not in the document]
Tanguylo: An error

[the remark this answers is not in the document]
Tanguylo: image

dessia_common/exports.py
Tanguylo: I write 
`line += ''.join([" ", self._string_in_table(self._value_to_str(value)), " |"])`
instead of
`line += " " + self._string_in_table(self._value_to_str(value)) + " |"`
which is not the same as suggested

dessia_common/exports.py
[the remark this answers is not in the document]
GhislainJ: Yes, indeed

dessia_common/exports.py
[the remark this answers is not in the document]
Tanguylo:
```
def _sequence_to_str(self, value: Sequence):
        if len(value) == 0:
            return f"empty {type(value).__name__}"

        printed_string = f"{len(value)} "

        all_class_names = list(set(subvalue.__class__.__name__ for subvalue in value))
        if len(all_class_names) == 1:
            printed_string += ''.join([f"{all_class_names[0]}",
                                       f"{'s' if len(value) > 1 else ''}"])
        else:
            str_all_class = str(all_class_names).translate(str(all_class_names).maketrans('', '', "{}'"))
            printed_string += f"{str_all_class}"

        return printed_string

    def _dict_to_str(self, value: Dict):
        return self._sequence_to_str(list(value.values()))
```
and 
```
def _value_to_str(self, value: Any) -> str:
        if isinstance(value, (float, int, bool, complex)):
            return str(round(value, 6))
        if isinstance(value, str):
            return (value if value != '' else 'no value')
        if is_sequence(value):
            return self._sequence_to_str(value)
        if isinstance(value, Dict):
            return self._dict_to_str(value)
        return self._object_to_str(value)
```

dessia_common/exports.py
Tanguylo: -1 allows me not to test if the value is None when i divide it by 2

dessia_common/exports.py
Tanguylo: Previously discussed, will not be implemented

dessia_common/exports.py
Tanguylo: Yes it does but I find it more readable

dessia_common/exports.py
Tanguylo: Yes it does but I find it more readable

dessia_common/templates.py
Tanguylo: mdr

dessia_common/forms.py
Tanguylo: I did as you asked

tests/markdowns.py
Tanguylo: That's right
I wrote some tests

masfaraud (approved):

dessia_common/exports.py
GhislainJ: Hmmm, even I am struggling right now to understand my previous comment. I think I mixed two ideas there.
I understand that joining rather than summing string is more efficient, but is this particular feature really that time consuming ? Otherwise, I don't think we should trade readability over performance.
I am suggesting something else, then, which looks way more correct than my first try : 
```
line = "|"
for stuff:
    line = line + f" {self._string_in_table(self._value_to_str(value))} |"
```

OR, if you really don't want to add strings, I suggest you build another list comprehension beforehand : 

```
all_strings = [self._string_in_table(self._value_to_str(v)) for v in row]
line = " | ".join(all_strings)
line = f"| {line} |" # Si toutefois il faut rajouter des | au début et à la fin. On peut faire ces deux dernières lignes en une seule, mais c'est plus dur à lire, je trouve
```

Which I think was my idea in the first place and would give, if I am not mistaken, the following result : 
`| awesome_string0 | awesome_string1 | ... | awesome_stringn |`
is it the intention behind it ?

dessia_common/exports.py
GhislainJ: This seems made up to me. It looks like you are tricking your way out of the "table_limit is -1" without exposing the true concept of the code here. If I understand correctly, by taking the -1 example, you are doing this : 
```
self.table_limit = -1
half_table = int(self.table_limite / 2) = int(-0.5) = 0
table = ""
content[:half_table] = content[:0] = []
for row in []: # is thus uneffective and table is not built

[...]

if len(content) > self.table_limit and self.table_limit != -1:
   # table limit is -1 so we donnot go in here

[...]

# The same goes for the second part with 
for row in content[-half_table:]:
    # We don't loop here because content[-half_table:] is []

return table # Which is ""
```

Is that the expected result ? If yes, I would totally expose the true concept of the table_limit not being defined, here, which is : "If table limit is None, then we do nothing and show an empty string", simply by doing : 
```
table = ""
if self.table_limit is None:
    return table
# Then,  all the other stuff as you already did : 
half_table = int(self.table_limit / 2)
if ...
```

And I think this is even more true if the case for no table_limit is not to be an empty string, but to be the whole formatted dataset. In this case you don't even need the half table stuff, so this is another algorithm

dessia_common/exports.py
GhislainJ: I guess we all have different sensibilities to it. In fact, one or two years ago, I would have totally agreed with you, but right now I find it totally barbaric. I get the fact that they are spaced into a column, though which make total sense

dessia_common/exports.py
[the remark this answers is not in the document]
GhislainJ: After reviewing this line again, why not keep the if as a one-liner ? ie
`if hasattr(value, 'name') and value.name: # (Perfectly ok to test a string a boolean for non-emptiness)`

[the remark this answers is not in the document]
Tanguylo: Thanks, the code is better now

dessia_common/exports.py
Tanguylo:
```
line = "|"
for stuff:
    line = line + f" {self._string_in_table(self._value_to_str(value))} |"
```

This one is very better

dessia_common/exports.py
Tanguylo: Changed the method following our oral conversation

GhislainJ (approved):

PR #403, opened 2022-10-24
fix(plot_dataset): change DataSet to Dataset

Tanguylo:
- [x] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:
Change DataSet to Dataset for plot_data


Tanguylo:
I compared the file vis.min.js in this PR with the same from master and there is no difference between the two of them

Tanguylo:
plot in forms.py is ok now

masfaraud (approved):

PR #421, opened 2022-11-08
fix: declare correct dependecy for sklearn

ladnane:
- [x] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [x] No

* **Other information**:
dependency on scikit learn declared in DC is not honored, this is breaking backend as it cause failed import of DC. I suggest this to be added as a patch as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -147,7 +147,7 @@ setup(
         "openpyxl",
         "parameterized",
         "matplotlib",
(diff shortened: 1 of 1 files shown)


Tanguylo (approved):
Required to use datatools

PR #436, opened 2022-11-28
Generate Empty Workflow

GhislainJ:
- [ ] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:


dessia_common/workflow/core.py
```diff
@@ -403,6 +399,14 @@ def __init__(self, blocks, pipes, output, *, imposed_variable_values=None,
         self.branch_by_display_selector = self.branch_by_selector(self.display_blocks)
         self.branch_by_export_format = self.branch_by_selector(self.export_blocks)
 
+    @classmethod
+    def generate_empty(cls):
+        return cls(blocks=[], pipes=[], output=None)
```
Tanguylo, line 404: Why not an __init__ with default values ?

Tanguylo (approved):
Maybe the generate_empty method should be stated with default values in init but it's not mandatory for merging in dev

PR #441, opened 2022-11-30
(2) chore(cluster): 100 percents coverage on cluster

Tanguylo:
- [ ] Bug fix
- [ ] new features
- [x] performance 
- [ ] docs update

- [ ] Yes:
- [x] No

* **Other information**:
Add tests for list methods


GhislainJ (approved):

dessia_common/exports.py
```diff
@@ -1,8 +1,5 @@
-#!/usr/bin/env python3
-# -*- coding: utf-8 -*-
```
masfaraud, line 2: Why remove encoding?

dessia_common/errors.py
```diff
@@ -1,13 +1,11 @@
-#!/usr/bin/env python3
```
masfaraud, line 1: why remove encoding?

masfaraud (commented):
Is this PR only doc? The title says it is for coverage but i don't see any test related stuff?

PR #447, opened 2022-12-01
(3) fix(markdown_writer): change value for none values as -

Tanguylo:
- [ ] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update
- [x] Improvement

- [ ] Yes:
- [x] No

* **Other information**:
Change displayed value in markdown table for None values and list of list


Tanguylo:
This PR makes the markdown table writer writing something in list and when value is None.
For None value, the str is '-'
For list, the str is now like that :
`str_value = str(len(list)) + ' [class_name_1, class_name_2, ..., class_name_n]'[:20]`

dessia_common/exports.py
```diff
@@ -317,6 +317,8 @@ def _value_to_str(self, value: Any) -> str:
             return self._sequence_to_str(value)
         if isinstance(value, Dict):
             return self._dict_to_str(value)
+        if value is None:
+            return '-'
```
GhislainJ, line 321: It looks like the returned value differs from the previous functions. The later had space around the ` - ` as this one doesn't. Is it intended ?

dessia_common/exports.py
```diff
@@ -327,7 +329,8 @@ def _object_matrix(self, object_):
         for attr, value in object_.__dict__.items():
             matrix.append([attr,
                            value.__class__.__name__,
-                           (self._value_to_str(value) if not isinstance(value, (list, dict, set)) else ' - ')])
+                           self._value_to_str(value)])
+                           # (self._value_to_str(value) if not isinstance(value, (list, dict, set)) else ' - ')])
```
GhislainJ, line 333: Is it important to keep this commented ?

GhislainJ (commented):
Some enquiries

dessia_common/exports.py
Tanguylo: Removed

dessia_common/exports.py
Tanguylo: I'm not sure it changes something for the final render of the markdown on web app but i surrounded the - with spaces, which probably wont make the difference but gives a more consistent print

PR #450, opened 2022-12-06
(2) Fix/doc dataset #449 + Docstring parsing #336 

AxelRinghausen:
- [ ] Bug fix
- [ ] new features
- [ ] performance 
- [x] docs update

- [ ] Yes:
- [x] No

* **Other information**:


dessia_common/forms.py
```diff
@@ -308,25 +308,24 @@ def generate_from_text(cls, stream: StringFile):
 
     def add_standalone_object(self, object_: StandaloneBuiltinsSubobject):
         """
-        Add a standalone object to object_list.
-
-        It doesn't return anything, hence, API will update object when computing from frontend.
+        Adds a standalone object to object_list.
```
GhislainJ, line 311: We discussed with Tanguy, and it looks like convention sais the verb should not have an "s" at the end

dessia_common/forms.py
```diff
@@ -1,6 +1,6 @@
 """
-Module that aims to list all possibilities of data formats offered by Dessia.
-
+A module that aims to list all
```
GhislainJ, line 2: Should this be a one-liner ?

dessia_common/exports.py
```diff
@@ -50,8 +51,9 @@ def __init__(self, selector: Optional[str], extension: str, method_name: str, te
         self.args = args
 
     def to_dict(self):
-        return {"selector": self.selector, "extension": self.extension, "method_name": self.method_name,
-                "text": self.text, "export_name": self.export_name, "args": self.args}
+        return {"selector": self.selector, "extension": self.extension,
```
GhislainJ, line 54: Do you still use the 80 char max line length rule ? We decided to set it to 120 now

GhislainJ (changes requested):

dessia_common/exports.py
AxelRinghausen: Yes I do.

GhislainJ (approved):

PR #451, opened 2022-12-06
Fix/workflow pipe order

GhislainJ:
- [ ] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:


dessia_common/workflow/blocks.py
```diff
@@ -67,18 +63,20 @@ def output_from_function(function, name: str = "result output"):
 
 
 class BlockError(Exception):
-    pass
+    """ Specific BlockError Exception. """
```
Tanguylo, line 66: pass is not required ?

dessia_common/workflow/core.py
```diff
@@ -257,15 +255,13 @@ def __init__(self, input_variable: Variable, output_variable: Variable, name: st
...
-        transform the pipe into a dict
-        """
+        """ Transform the pipe into a dict. """
         return {'input_variable': self.input_variable, 'output_variable': self.output_variable,
                 'memorize': self.memorize}
 
 
 class WorkflowError(Exception):
-    pass
+    """ Specific WorkflowError Exception. """
```
Tanguylo, line 264: pass ?

dessia_common/workflow/core.py
```diff
@@ -508,54 +504,51 @@ def __deepcopy__(self, memo=None):
...
+        This needs special care because if it is not handled properly, NBVs can duplicate,
+        or copied pipes might be unordered.
+        """
+        nbv = pipe.input_variable
+        upstream_index = self.variable_index(nbv)
+        if upstream_index in copy_memo:
+            copied_variable = copy_memo[upstream_index]
+        else:
+            copied_variable = nbv.copy()
+            copy_memo[upstream_index] = copied_variable
```
Tanguylo, line 531: can upstream_index be None ?
If yes, what if no value in upstream_index ?

dessia_common/workflow/core.py
```diff
@@ -1039,6 +1014,9 @@ def variable_indices(self, variable: Variable) -> Optional[Union[Tuple[int, int,
             return self.nonblock_variables.index(upstream_variable)
         raise WorkflowError(f"Something is wrong with variable {variable.name}")
 
+    def is_variable_nbv(self, variable: Variable) -> bool:
+        return isinstance(self.variable_indices(variable), int)
```
Tanguylo, line 1018: Are you sure this test is enough ?

Tanguylo (approved):

dessia_common/workflow/blocks.py
GhislainJ: Well, I was a bit confused as well, but I got yelled at by pylint because of "unecessary pass". So I guess it is not if you set a docstring, which is pretty cool in my opinion

dessia_common/workflow/core.py
GhislainJ: Yes, that is totally sufficient.
Actually, NBVs are some kind of weird objects, between variables and blocks.
They are variables, but variables should be attached to blocks. In this case, we can give it an adress that is `(i_block, i_side, i_port)`. If it isn't attached to a block, well, it kinda behaves like a block and only have an int as an "adress". That is why we can spot it just by testing it is an instance of int.

Any other kind of value here will be faulty anyway

dessia_common/workflow/core.py
GhislainJ: I'll have to give it a second thought. I see no way it could be None, but now that you mention it, I have to ensure that it cannot

dessia_common/workflow/blocks.py
Tanguylo: Yes that's nice !
Pylint is a nice teacher !

PR #455, opened 2022-12-12
Reference path

GhislainJ:
- [ ] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:


ladnane (changes requested):
Hello Ghislain,
Can you create/link an issue, detailling what the current PR is adressing.
I am probably wrong, but am afraid the direction of these changes is not in accordance with the changes what we discussed in term of moving toward an Id based ref "path" wich the backend would be able to serve directly ?!
That would help anticipate any impact on backend side as well.
thanks

GhislainJ:
This is not directly linked with "IDfied" references, as it is only used in the scope of displays. It basically is a path created by the user to reference the object that is represented by a 3D part or a plot_data point, for instance.

In fact, this construction has no way to understand what would be an "ID reference" globally created at object level, as this only deals with "giving a path to frontend in order to get any more info".

Frontend side, it will be used as such : 
- Read the json & build the view,
- On some user input (selection, etc), try and get the "reference_path" defined in the json which denotes "how to recursively get to subobject" (#/path/to/subobject), 
- If this path is understandable, ask backend to get it

Backend side, then,  the only location where this would be useful (for now), would be in the object route when it is called with a subobject path. 
I guess, from a dc (displays, not serialization) and frontend perspective, that your modifications should be seen a refactors and not require any changes on frontend side. This means that we would still call the object route (or a new one, specifically to get subobjects, even) with a path that looks like so (#/path/to/subobject), because we are totally agnostic on the context (frontend would not know how to call an ID ref, because it doesn't know what object parenthood is).
In this case, I also guess, that this **should** not impact your devs as long as you have one or several functions able to perform a complex get (ref or refs) from a complete path.
Am I wrong on this whole last part ?

ladnane (approved):

masfaraud (approved):

PR #481, opened 2023-01-08
save workflow_run to python script

younesdessia:
- [ ] Bug fix
- [X] new features
- [ ] performance 
- [ ] docs update

- [X] Yes:
- [ ] No

* **Other information**:

Add methods to save workflow run to python script (py)


dessia_common/workflow/core.py
```diff
@@ -2186,6 +2186,44 @@ def _method_jsonschemas(self):
...
+        """
+        # TODO : don't take into account yet TypedVariable.
+        # TODO : take into account the type of inputs (type_)
+        workflow_script = self.workflow.to_script()
+        input_str = ""
+        default_value = ""
+        for j, block in enumerate(self.workflow.blocks):
+            for i, input_ in enumerate(block.inputs):
+                if not input_.has_default_value and not \
+                        'block_' + str(j) + '.inputs' + '[' + str(i) + ']' in workflow_script:
```
GhislainJ, line 2201: As it is used twice, `'block_' + str(j) + '.inputs' + '[' + str(i) + ']'` should be set in a variable.

Consider using fstring as well

dessia_common/workflow/core.py
```diff
@@ -2186,6 +2186,44 @@ def _method_jsonschemas(self):
...
+    def to_script(self):
+        """
+        Computes a script representing the workflowrun.
+        """
+        # TODO : don't take into account yet TypedVariable.
+        # TODO : take into account the type of inputs (type_)
+        workflow_script = self.workflow.to_script()
+        input_str = ""
+        default_value = ""
+        for j, block in enumerate(self.workflow.blocks):
```
GhislainJ, line 2198: WorkflowRun knows its input_values. Couldn't we use this, instead of looping on both blocks and inputs ?

GhislainJ (changes requested):
I foresee a limitation here. This will be difficult to use with streams and files input locally, as file location are always different depending on which user is running the script.

Anyway, the PR lacks two main things, in addition to the comments made to the code : 
- Changelog modification
- Tests in order to check the feature

dessia_common/workflow/core.py
younesdessia: @GhislainJ 
it's ok i changed that

dessia_common/workflow/core.py
```diff
@@ -2186,6 +2186,67 @@ def _method_jsonschemas(self):
...
+        Computes a script representing the workflowrun.
+        """
+        # TODO : don't take into account yet TypedVariable.
+        # TODO : take into account the type of inputs (type_)
+        workflow_script = self.workflow.to_script()
+        workflow_script_import = self.workflow._to_script().imports
+        input_str = ""
+        default_value = ""
+        add_import = ""
+        var_1 = "block_{}.inputs[{}]"
```
GhislainJ, line 2200: We should set this variable inside the loop to avoid the .format() stuff and use fstrings instead.

Also var_1 is a poor naming and should be improved. Why not variable or even just var ?

dessia_common/workflow/core.py
```diff
@@ -2186,6 +2186,67 @@ def _method_jsonschemas(self):
...
+        for j, block in enumerate(self.workflow.blocks):
+            for i, input_ in enumerate(block.inputs):
+                if not input_.has_default_value and not \
+                        var_1.format(j, i) in workflow_script:
+                    input_str += f"    workflow.input_index(" \
+                                 f"{var_1.format(j, i)}):" \
+                                 f" value_{str(j) + '_' + str(i)},\n"
+                    default_value_ = f"\nvalue_{j}_{i} = 0"
```
GhislainJ, line 2208: Why is default value set to 0 ?

dessia_common/workflow/core.py
```diff
@@ -2186,6 +2186,67 @@ def _method_jsonschemas(self):
...
+                    if not isinstance(input_, (str, float, int)):
+                        try:
+                            module_ = input_._get_to_script_elements().get_import_dict()
+                            key, value = list(module_.items())[0]
```
GhislainJ, line 2212: Dangerous indexing of `items` function

dessia_common/workflow/core.py
```diff
@@ -2186,6 +2186,67 @@ def _method_jsonschemas(self):
...
+                            default_value_ = f"\nvalue_{j}_{i} = {value[0]}()"
+                        except:
+                            pass
+                        if not any(value[0] in lst for lst in (add_import, workflow_script_import)):
+                            add_import += f"from {key} import {value[0]}\n"
+                        default_value += default_value_
+        for k, nbv in enumerate(self.workflow.nonblock_variables):
+            if not nbv.has_default_value:
```
GhislainJ, line 2220: Why do we set default values if variables should not be defaulted ?

dessia_common/workflow/core.py
```diff
@@ -2186,6 +2186,67 @@ def _method_jsonschemas(self):
...
+                if nbv.type_ == str:
```
GhislainJ, line 2221: Should use isinstance for these checks

dessia_common/workflow/core.py
```diff
@@ -2186,6 +2186,67 @@ def _method_jsonschemas(self):
...
+                    default_value += f"\nvalue_nbv_{k} = 'str'"
+                if nbv.type_ == int:
+                    default_value += f"\nvalue_nbv_{k} = 'int'"
+                input_str += f"    workflow.input_index(" \
```
GhislainJ, line 2225: Instead of writing input_index into the string, we could compute the right value for the index beforehand and write `f"{{global_index} : value_nbv_{k}}`. 

You can also get rid of the `str()` in fstrings

Be aware of the style conventions in computed code (spaces around " : ", space only after ", ",...)

dessia_common/workflow/core.py
```diff
@@ -2186,6 +2186,67 @@ def _method_jsonschemas(self):
...
+                    input_str += f"    workflow.input_index(" \
+                                 f"{var_1.format(j, i)}):" \
+                                 f" value_{str(j) + '_' + str(i)},\n"
+                    default_value_ = f"\nvalue_{j}_{i} = 0"
+                    if not isinstance(input_, (str, float, int)):
+                        try:
+                            module_ = input_._get_to_script_elements().get_import_dict()
+                            key, value = list(module_.items())[0]
+                            default_value_ = f"\nvalue_{j}_{i} = {value[0]}()"
+                        except:
```
GhislainJ, line 2214: Dangerous bare except

dessia_common/workflow/core.py
```diff
@@ -2186,6 +2186,67 @@ def _method_jsonschemas(self):
...
+                            pass
+                        if not any(value[0] in lst for lst in (add_import, workflow_script_import)):
```
GhislainJ, line 2216: What the check do here isn't clear, we could set it in a variable in order for the if statement to be "sentence-like".

in addition to that, I am not a big fan of the "lst" variable name. What does it represent ?

GhislainJ (changes requested):

dessia_common/workflow/core.py
younesdessia: Here we check if the class (module) of the input is not imported before

dessia_common/workflow/core.py
younesdessia: No Here I can't use isinstance to check the type of nbv

dessia_common/workflow/core.py
younesdessia: Here it's not the defaut variable of input, but the value I will give to the input without default value

PR #483, opened 2023-01-09
(2) fix(refpath_dataset): implements reference path for dataset and clust…

Tanguylo:
- [x] Bug fix
- [x] new features

- [ ] Yes:
- [x] No

* **Other information**:
Implements reference path for dataset and clustereddataset


dessia_common/datatools/cluster.py
```diff
@@ -315,12 +315,13 @@ def plot_data(self):
...
-        for row, label in enumerate(self.labels):
-            _plot_data_list.append({attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)})
-            _plot_data_list[-1]["Cluster Label"] = label
-            # (label if label != -1 else "Excluded") plot_data "Excluded" -> NaN
-        return _plot_data_list
+        plot_data_list = []
+        for row, (dobject, label) in enumerate(zip(self.dessia_objects, self.labels)):
+            sample_values = {attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)}
+            sample_values["Cluster Label"] = label
+            reference_path = f"dessia_objects/{row}"
```
GhislainJ, line 322: Should add prefix "#/"

dessia_common/datatools/cluster.py
```diff
@@ -315,12 +315,13 @@ def plot_data(self):
         return Dataset.plot_data(self)
 
     def _plot_data_list(self):
-        _plot_data_list = []
-        for row, label in enumerate(self.labels):
-            _plot_data_list.append({attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)})
-            _plot_data_list[-1]["Cluster Label"] = label
-            # (label if label != -1 else "Excluded") plot_data "Excluded" -> NaN
-        return _plot_data_list
+        plot_data_list = []
```
GhislainJ, line 318: `samples` would be a pretty neat name for this in my opinion (follows conventions, is consistent with sample_values defined afterwards, is straight to the point)

dessia_common/datatools/cluster.py
```diff
@@ -315,12 +315,13 @@ def plot_data(self):
...
+        for row, (dobject, label) in enumerate(zip(self.dessia_objects, self.labels)):
```
GhislainJ, line 319: I find `dobject` naming a bit poor as it is an abbreviation. Why not dessia_object or even just object_ ? (trailing underscore for escaping python-reserved "object" global variable)

dessia_common/datatools/cluster.py
```diff
@@ -315,12 +315,13 @@ def plot_data(self):
...
+            sample_values = {attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)}
+            sample_values["Cluster Label"] = label
+            reference_path = f"dessia_objects/{row}"
+            plot_data_list.append(Sample(sample_values, reference_path, dobject.name))
```
GhislainJ, line 323: I am wondering what would be the input if none of the dessia_objects are named. Maybe we could default it in that case ? For ex : 
```py
name = dessia_object.name
if not name:
    name = f"Sample {row}"
```

dessia_common/datatools/dataset.py
```diff
@@ -671,8 +671,10 @@ def _tooltip_attributes(self):
 
     def _plot_data_list(self):
```
GhislainJ, line 672: Same remarks than `cluster` here

GhislainJ: Well, on second thought, this be of code could even be mutualized between the two of them

GhislainJ (changes requested):

dessia_common/datatools/cluster.py
Tanguylo: Done

dessia_common/datatools/cluster.py
Tanguylo: Done

dessia_common/datatools/cluster.py
Tanguylo: Done

dessia_common/datatools/cluster.py
Tanguylo: Done

dessia_common/datatools/dataset.py
Tanguylo: Done

dessia_common/datatools/dataset.py
```diff
@@ -669,11 +669,18 @@ def _histogram_unic_value(self, idx_col: int, name_attr: str):
     def _tooltip_attributes(self):
         return self.common_attributes
 
-    def _plot_data_list(self):
-        plot_data_list = []
-        for row, _ in enumerate(self.dessia_objects):
-            plot_data_list.append({attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)})
-        return plot_data_list
+    def _object_to_sample(self, dessia_object: DessiaObject, row: int):
```
GhislainJ, line 672: I would go even further that that. Consider this : 

```python
def _object_to_sample(self, dessia_object: DessiaObject, row: int, label: str = None) -> Sample:
    sample_values = {attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)}
    if label is not None:
        sample_values["Cluster Label"] = label
        reference_path = f"#/dessia_objects/{row}"
        name = dessia_object.name if dessia_object.name else f"Sample {row}"
    return Sample(values=sample_values, reference_path=reference_path, name=name)

# DATASET
def _to_samples(self):
    return [self._object_to_sample(dessia_object=object_, row=row)
            for row, object_ in enumerate(self.dessia_objects)]
    
# CLUSTER
def _to_samples(self):
     return [self._object_to_sample(dessia_object=object_, row=row, label=label)
             for row, (object_, label) in enumerate(zip(self.dessia_objects, self.labels))]
```

Notes : 
- Normally, I would have called loop variables inside comprehension list as one letters, withtout naming function arguments like this : 
`samples = [object_to_samples(o, i, l) for i, (o, l) in enumerate(zip(self.objects, self.labels))]`
BUT, I am annoyed here, because "l" is forbidden as a one-letter variable name, as it is easily confused with a 1 or a I depending on the font, so that is why I named them without abbreviating AND I named the arguments for clarity (it is "consistence of the length" in a way, you either go full short or full complete)

- When you have more than one arg in a function call or an object instantiation, prefer naming args in the statement. For ex :
` Sample(values=sample_values, reference_path=reference_path, name=name)`. this is because it is easier to read and understand and is robust to any refactor. If we add, remove or change orders of arguments, the code will still work as long as there weren't any breaking changes. 
BUT not when there is only one, because it is normally straightforward AND it can cause a bug when the code is compiled with Cython. ex : 
`object_._display_from_selector(selector)`
and it is not that subject to refactor as there is only one arg

GhislainJ (changes requested):

dessia_common/datatools/dataset.py
GhislainJ: Note, that if you are inheriting your clusters with Dataset, you could, in my code, instead of giving a label arg to the `_object_to_sample` method that is never used in the case of the Dataset, make use of the inheritance. What I mean is do the following : 

- Define your base `_object_to_sample` and `_to_samples` method in Dataset
- Make Cluster inherit from Dataset
- Overwrite `_object_to_sample` method
- Inside, call the super() `_object_to_sample` method,
- Loop on a zip of it and self.labels, in order to change already created values
- Never overwrite `_to_samples` method

That might be the right way to do it

GhislainJ: Like this : 
```python

class Dataset:
    def __init__(self):
        self.common_attributes = []
        self.matrix = [[]]
        self.dessia_objects = []
        self.labels = []
        pass

    def _object_to_sample(self, dessia_object: DessiaObject, row: int) -> Sample:
        sample_values = {attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)}
        reference_path = f"#/dessia_objects/{row}"
        name = dessia_object.name if dessia_object.name else f"Sample {row}"
        return Sample(values=sample_values, reference_path=reference_path, name=name)

    # DATASET
    def _to_samples(self):
        return [self._object_to_sample(dessia_object=object_, row=row)
                for row, object_ in enumerate(self.dessia_objects)]


class Cluster(Dataset):
    def __init__(self):
        Dataset.__init__(self)

    def _object_to_sample(self, dessia_object: DessiaObject, row: int) -> Sample:
        samples = super(Cluster, self)._object_to_sample(dessia_object=dessia_object, row=row)
        for sample, label in zip(samples, self.labels):
            sample.values["Cluster Label"] = label
        return samples
```

Tanguylo: I propose to make approximately what is proposed in the first idea. Without rewriting _to_sample method.

dessia_common/datatools/dataset.py
```diff
@@ -666,11 +663,20 @@ def _histogram_unic_value(self, idx_col: int, name_attr: str):
     def _tooltip_attributes(self):
         return self.common_attributes
 
-    def _plot_data_list(self):
-        plot_data_list = []
-        for row, _ in enumerate(self.dessia_objects):
-            plot_data_list.append({attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)})
-        return plot_data_list
+    def _object_to_sample(self, dessia_object: DessiaObject, reference_path: str, row: int):
```
GhislainJ, line 666: It would really be a better code design if object_ to_sample method returned a Sample object, like proposed in my previous comment.

This would : 
- be easier to maintain, because hauling variables through functions and returns is cumbersome,
- thus, be more "Object-Oriented-like" or "pythonic",
- allow a way simpler _to_samples method that could even be written as a one-liner
- globally, be more understandable by any other developper reading this bit of code

dessia_common/datatools/dataset.py
```diff
@@ -666,11 +663,20 @@ def _histogram_unic_value(self, idx_col: int, name_attr: str):
...
+        sample_values = {attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)}
+        full_reference_path = f"{reference_path}/dessia_objects/{row}"
+        name = dessia_object.name if dessia_object.name else f"Sample {row}"
+        return sample_values, full_reference_path, name
+
+    def _to_samples(self, reference_path: str):
```
GhislainJ, line 672: reference_path arg could (should ?) be defaulted as "#" here as well. It is also the case in the object_to_sample method. In the case anyone would inherit your class and use these methods, it would allow them to use the feature by default and not break the behavior in their object

dessia_common/datatools/dataset.py
```diff
@@ -666,11 +663,20 @@ def _histogram_unic_value(self, idx_col: int, name_attr: str):
...
         return self.common_attributes
 
-    def _plot_data_list(self):
-        plot_data_list = []
-        for row, _ in enumerate(self.dessia_objects):
-            plot_data_list.append({attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)})
-        return plot_data_list
+    def _object_to_sample(self, dessia_object: DessiaObject, reference_path: str, row: int):
+        sample_values = {attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)}
+        full_reference_path = f"{reference_path}/dessia_objects/{row}"
```
GhislainJ, line 668: Don't be afraid to overwrite reference_path name, here. that's what I would do, instead of defining a new variable

GhislainJ (changes requested):

dessia_common/datatools/dataset.py
Tanguylo: Actually, reference_path is a sequence and when you rewrite it, it behaves like an empty list:
`reference_path = f"{reference_path}/dessia_objects/{row}"` is taking previously created reference path to update it... so you get a "dessia_objects/0/dessia_objects/1/dessia_objects/2/..." reference path

dessia_common/datatools/dataset.py
Tanguylo: I was thinking that if plot_data gives a value to reference_path, these methods don't need one but if you tell me the opposite, I put this default value

dessia_common/datatools/dataset.py
Tanguylo: Done.
Actually it was a little uggly.

Be careful, i write on values attribute of sample in clustereddataset to handle the label of clusters in plot method

dessia_common/datatools/dataset.py
Tanguylo: Well, i did it

GhislainJ: The thing is, someone could use these without going through plot_data

dessia_common/datatools/dataset.py
GhislainJ: I think it is because the old version of your function (that did not return a sample). I am pretty sure the reference_path value did not behave like a sequence (well, it is, but a str, so its immutable), but, as you built it, returned it and reuse in the for loop in your `to_samples` method, it got updated every step in the loop. 

With the new way of defining your function, you should not have the same behavior, because reference_path's modification scope keeps being restricted to the `object_to_sample` method

GhislainJ (changes requested):

dessia_common/datatools/cluster.py
```diff
@@ -305,13 +305,10 @@ def plot_data(self, reference_path: str = "#", **kwargs):
...
 
-    def _plot_data_list(self):
-        _plot_data_list = []
-        for row, label in enumerate(self.labels):
-            _plot_data_list.append({attr: self.matrix[row][col] for col, attr in enumerate(self.common_attributes)})
-            _plot_data_list[-1]["Cluster Label"] = label
-            # (label if label != -1 else "Excluded") plot_data "Excluded" -> NaN
-        return _plot_data_list
+    def _object_to_sample(self, dessia_object: DessiaObject, row: int, reference_path: str = '#'):
+        new_sample = super()._object_to_sample(dessia_object=dessia_object, row=row, reference_path=reference_path)
```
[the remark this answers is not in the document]
Tanguylo: done

GhislainJ (approved):

PR #501, opened 2023-01-19
Fix nbv to_script

younesdessia:
- [X] Bug fix

- [ ] Yes:
- [x] No

* **Other information**:
this PR fix to_script, if we have nbv which contains a list of objects


GhislainJ (changes requested):
As stated in PM, this need to be changed be fore being merged

dessia_common/workflow/core.py
```diff
@@ -101,15 +101,44 @@ def dict_to_object(cls, dict_: JsonSerializable, force_generic: bool = False,
...
+        Update elements of the script representing the workflow.
+
+        return ToSriptElement
+        """
+        types = ["Dict", "Tuple"]
+        try:
+            script.declaration += f", type_={self.type_.__name__}"
+        except AttributeError:
+            if is_typing(self.type_):
+                args = self.type_.__args__
```
GhislainJ, line 116: You should use `get_args` from `typing` module

dessia_common/workflow/core.py
```diff
@@ -101,15 +101,44 @@ def dict_to_object(cls, dict_: JsonSerializable, force_generic: bool = False,
...
+                if len(args) == 1:
```
GhislainJ, line 117: This is not generic. typings could have any number of args, so we cannot only have two cases for 1 and 2.

You have to forge the str from any number of args

dessia_common/workflow/core.py
```diff
@@ -101,15 +101,44 @@ def dict_to_object(cls, dict_: JsonSerializable, force_generic: bool = False,
...
+                    args = self.type_.__args__[0].__name__
+                elif len(args) == 2:
+                    args = f"{args[0].__name__}, {args[0].__name__}"
+                script.declaration += f", type_={serialize_typing(self.type_).split('[')[0]}" \
+                                      f"[{args}]"
+                script.imports.append(f"{self.type_.__args__[0].__module__}.{self.type_.__args__[0].__name__}")
```
GhislainJ, line 123: I don't understand this. Why do we take only index 0 ?

dessia_common/workflow/core.py
```diff
@@ -101,15 +101,44 @@ def dict_to_object(cls, dict_: JsonSerializable, force_generic: bool = False,
...
+        if "builtins" not in serialize_typing(self.type_) or any(type_ in serialize_typing(self.type_)
```
GhislainJ, line 124: I don't understand this statement as well. Wy do we search builtins not in the serialized type ? Why do we check for only Dict or Tuple in it ?

dessia_common/workflow/core.py
```diff
@@ -101,15 +101,44 @@ def dict_to_object(cls, dict_: JsonSerializable, force_generic: bool = False,
...
+                                                                 for type_ in types):
+            try:
+                script.imports.append(f"{self.type_.__module__}.{(self.type_.__origin__.__name__.capitalize())}")
```
GhislainJ, line 127: You should use get_origin from typing module

dessia_common/workflow/core.py
```diff
@@ -101,15 +101,44 @@ def dict_to_object(cls, dict_: JsonSerializable, force_generic: bool = False,
...
+            except AttributeError:
```
GhislainJ, line 128: For which case stands the AttributeError ?

GhislainJ (changes requested):
In order to be better understood, you should add tests cases. We could use them as support to know what we are actually doing here

dessia_common/workflow/core.py
younesdessia: because `self.type__.__args__ ` is a tuple (ex :` (dessia_common.files.BinaryFile,)`)

dessia_common/workflow/core.py
younesdessia: it's true, 
I did that because in the case of a dict we don't have just one args,

dessia_common/workflow/core.py
younesdessia: because when we do `serialize typing(self.type) `on a Dict we get `'Dict[__builtins__.str, __builtins__.str]'`, it does not fit in the condition

dessia_common/workflow/core.py
younesdessia: ok i will try that

dessia_common/workflow/core.py
younesdessia: in the case where object has no attribute '__origin__', 
exemple : `dessia_common.datatools.dataset.Dataset`

dessia_common/workflow/core.py
GhislainJ: Yes, from my experience, you have these cases : 
- List, InstanceOf, Subclass, Type should have only one arg (exactly 1),
- Dict should have two args (exactly 2),
- Tuple could have any number of args, but at least 1,

You have weird typings on whoch we don't have the hand on. For example if you type this : 
`def function(arg: T = None):`
Then, you have the following type : 
`Optional[T]`
Which have two args, the first one is T, the second one is NoneType. And this is only true when the value is defaulted to None.

We may have more cases, but these are the only one I can think of right now

dessia_common/workflow/core.py
GhislainJ: That is why you may want to use the is_typing function (from dessia_common.utils.types). We consider a typing anything that has `__origin__`, `__args__` & `__module__`.

Custom classes are not considered as typings, that is why we now try to to clearly state the differences between annotation and typing

PR #520, opened 2023-02-07
Fix to dict

masfaraud:
- [ ] Bug fix
- [ ] new features
- [X] performance: avoid overserializing non standalone objects 
- [ ] docs update

- [ ] Yes:
- [X] No

* **Other information**:

### DIFF
--- a/dessia_common/serialization.py
+++ b/dessia_common/serialization.py
@@ -26,6 +26,7 @@ fullargsspec_cache = {}
 class SerializableObject(CoreDessiaObject):
     """ Serialization capabilities of Dessia Object. """
 
+    _standalone_in_db = False
     _non_serializable_attributes = []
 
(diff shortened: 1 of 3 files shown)


dessia_common/serialization.py
```diff
@@ -160,7 +161,8 @@ def serialize_with_pointers(value, memo=None, path='#', id_method=True, id_memo=
             id_ = str(uuid.uuid1())
             id_memo[id_] = serialized
             memo[value] = f'#/_references/{id_}'
```
ladnane, line 163: I am not sure  this is enough to resolve the issue. If my understunding is correct, this will still add a new reference in _references, on backend side, we can't tell if this ref is for a standalone obj or not ( unless we do some code inspection i guess) and so it will be saved as a separate entry in db.

PR #522, opened 2023-02-08
fix(str_dataset): print dataset with values that are not attributes

Tanguylo:
- [x] Bug fix

- [ ] Yes:
- [x] No

* **Other information**:
Allow to print Dataset with values that are not attributes


dessia_common/datatools/cluster.py
```diff
@@ -97,10 +97,10 @@ def _write_str_prefix(self):
         prefix += (f"{len(self)} samples, {len(self.common_attributes)} features, {self.n_clusters} clusters")
         return prefix
 
-    def _get_printed_value(self, dessia_object: DessiaObject, attr: str):
+    def _get_printed_value(self, dessia_object_idx: int, attr: str):
```
GhislainJ, line 100: In order to avoid abbrevation in variable names, consider naming the argument `object_index` or even just `index`

tests/dataset.py
```diff
@@ -51,6 +51,7 @@ def to_vector(self):
 bidon = Bidon()
 bidon_hlist = Dataset([bidon] * 10)
 bidon_hlist.plot_data()
+print(bidon_hlist)
```
GhislainJ, line 54: Has this print been let there on purpose or has it been forgotten ?

dessia_common/datatools/dataset.py
```diff
@@ -273,8 +273,12 @@ def to_markdown(self) -> str:
         return templates.dataset_markdown_template.substitute(name=name, class_=class_, element_details=element_details,
                                                               table=table)
 
-    def _get_printed_value(self, dessia_object: DessiaObject, attr: str):
-        return getattr(dessia_object, attr)
+    def _get_printed_value(self, dessia_object_idx: int, attr: str):
+        # return getattr(dessia_object, attr)
```
GhislainJ, line 277: Could remove forgotten commented previous code

dessia_common/datatools/dataset.py
```diff
@@ -210,9 +210,9 @@ def _printed_attributes(self):
 
     def _print_objects_slice(self, key: slice, attr_space: int):
         string = ""
-        for dessia_object in self.dessia_objects[key]:
+        for dessia_object_idx in range(len(self[key])):
```
GhislainJ, line 213: I think just `i` would be justified here as a loop variable, since we loop in a range. It would be less cumbersome, in my opinion.

As a sidenote, indexing self bugs me out a bit. I don't find any reason why it would be bad, as long as self behaves as a sequence, but I am a bit puzzled by it and find it weird. Do not change anything for now, because I think this is OK, but we might consider explicitly using `__getitem__` in the future

GhislainJ (changes requested):
As a general comment and before making any changes, if this is a bug fix, could it be made on master or testing ? Or is the bug only on dev, right now ?

dessia_common/datatools/dataset.py
Tanguylo: I guess pylint is not happy when using __getitem__ instead of self indexing, because it is not happy when you write self.__len__(), saying that you should write len(self).

tests/dataset.py
Tanguylo: It is here on purpose but maybe you would like me to write an assertion ?

dessia_common/datatools/cluster.py
Tanguylo: OK, done.
I kept index instead of i (that you advised) in the for loop

dessia_common/datatools/dataset.py
Tanguylo: Done

Tanguylo:
> As a general comment and before making any changes, if this is a bug fix, could it be made on master or testing ? Or is the bug only on dev, right now ?

I changed for testing which contains the devs required for this bug resolution.

Master could take this also, but the __str__ method is not the same from master to testing so i prefer letting it on testing.

tests/dataset.py
GhislainJ: Fully asserting string equalities can be a bit overkill. Why not asserting the good length of the string ?

If the print is for checking purpose, chances are it will never be manually checked in CI process, so we should assert something, indeed

dessia_common/datatools/dataset.py
GhislainJ: Ok, for these kind of cases we can let pylint complain or not watch the error, but we will see this later

tests/dataset.py
Tanguylo: I made a test

GhislainJ (approved):

GhislainJ (approved):

PR #531, opened 2023-02-20
fix: to_dict fix for non standalone

masfaraud:
- [x] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:


dessia_common/serialization.py
```diff
@@ -151,28 +151,28 @@ def serialize_with_pointers(value, memo=None, path='#', id_method=True, id_memo=
 
     if isinstance(value, SerializableObject):
         if value in memo:
-            path_value, serialized_value, id_ = memo[value]
+            path_value, serialized_value, id_, _ = memo[value]
```
GhislainJ, line 154: Maybe we should create a Memo object instead of stating an arbitrary tuple

dessia_common/serialization.py
```diff
@@ -253,11 +251,21 @@ def serialize_dict_with_pointers(dict_, memo, path, id_method, id_memo):
...
 
     if path == '#':
-        # adding _references
-        serialized_dict['_references'] = id_memo
+        add_references(serialized_dict, memo, id_memo)
     return serialized_dict, memo
 
 
+def add_references(dict_, memo, id_memo):
+    # adding _references
```
GhislainJ, line 259: Comment redundant with function name

PR #542, opened 2023-02-24
perf: remove name from dict if empty

masfaraud:
- [x] performance: remove name from dict if empty

- [ ] Yes:
- [x] No

* **Other information**:

### DIFF
--- a/dessia_common/core.py
+++ b/dessia_common/core.py
@@ -125,7 +125,17 @@ class DessiaObject(SerializableObject):
     def base_dict(self):
         """ Base dict of the object, with just its name. """
         dict_ = SerializableObject.base_dict(self)
-        dict_['name'] = self.name
+        return dict_
+
+    def _serializable_dict(self):
+        """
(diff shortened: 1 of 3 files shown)


dessia_common/core.py
```diff
@@ -125,7 +125,17 @@ def __init__(self, name: str = '', **kwargs):
     def base_dict(self):
         """ Base dict of the object, with just its name. """
         dict_ = SerializableObject.base_dict(self)
-        dict_['name'] = self.name
+        return dict_
+
+    def _serializable_dict(self):
```
GhislainJ, line 130: Why not injecting name in Serializable object instances ? Currently, classes who are going to inherit from Serializable won't benefit from this change.

Also, it might have implications on frontend, as I don't know what happened if name is undefined (and not defined with empty string)

masfaraud: name is not an attribute of serializable object

GhislainJ: That is why I commented. Should we add it to SerializableObject ?

PR #544, opened 2023-02-27
perf: replace $ref with dc__ref during serialization

ladnane:
- [ ] Bug fix
- [ ] new features
- [x] performance 
- [ ] docs update

- [x] Yes:
- [] No

* **Other information**:

It would be nice to avoid using $ref in DC serialization since this is reserved keyword in mongodb.
This would allow removing a trasformation needed on backend side each time an object is saved or retrieved from db improving perf and reducing code complexity.


GhislainJ:
@ladnane I am ok with the idea. However, is the name `dc__ref` a requirement for the backend ? Maybe we should find a more convenient name for this, as it should not be mixed with a variable name that a user would have chosen for an object attribute.

In addition to that, did you witness the fact that the old objects still work on a global platform scale ? Our checks might be a bit poor to ensure this, currently, even if I think CI should have been broken if it didn't

ladnane:
@GhislainJ `dc__ref` is not a requirement, it's the key used on backend side, am ok with any other key. 
Unfortunatly I cannot check every object on every platform however i can garanty [through a backend migration] that those who use DC serialization or at least do not hardcode `$ref` values in their serialization should not have any issue during upgrade.

masfaraud (changes requested):
Change to ref__

PR #551, opened 2023-03-01
Deprecate Jsonschema

GhislainJ:
- [ ] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:


dessia_common/core.py
```diff
@@ -182,138 +164,45 @@ def schema(cls):
...
+        warnings.warn("'jsonschema' method is deprecated. Use schema instead", DeprecationWarning)
+        return cls.schema()
 
     @property
-    def method_schemas(self):  # TODO This should be a classmethod, but is a property to avoid collision with workflow.
+    def raw_method_schemas(self):
         """ Generate dynamic schemas for methods of class. """
+        # TODO This should be a classmethod, but is a property to avoid collision with workflow.
         cls = self.__class__
         valid_method_names = [m for m in dir(cls) if not m.startswith('_') and m in cls._allowed_methods]
```
Tanguylo, line 183: Are you sure _allowed_methods are always specified ? I guess, if not, all methods are allowed unless they have a specific pattern

dessia_common/schemas/core.py
```diff
@@ -442,8 +480,9 @@ def __init__(self, annotation: Type[T], attribute: str, definition_default: T =
     def to_dict(self, title: str = "", editable: bool = False, description: str = ""):
         """ Write Optional as a Dict. """
         default_value = self.schema.default_value()
-        chunk = self.schema.to_dict(title=title, editable=editable, description=description)
-        chunk["default_value"] = default_value
+
+        base_chunk = self.schema.to_dict(title=title, editable=editable, description=description)
+        chunk = {**base_chunk, "default_value": default_value}
```
Tanguylo, line 485: Elegant !

dessia_common/schemas/core.py
```diff
@@ -598,11 +637,15 @@ def serialized(self) -> str:
...
+    def standalone_in_db(self) -> bool:
+        """ Whether the class is standalone in db. """
+        return self.annotation._standalone_in_db
+
     def to_dict(self, title: str = "", editable: bool = False, description: str = ""):
         """ Write CustomClass as a Dict. """
         chunk = super().to_dict(title=title, editable=editable, description=description)
-        chunk.update({'type': 'object', 'standalone_in_db': self.annotation._standalone_in_db,
-                      "classes": [self.serialized]})
+        chunk.update({'type': 'object', 'standalone_in_db': self.standalone_in_db, "classes": [self.serialized]})
```
Tanguylo, line 648: Are you sure you want to compute twice (one wrong and one good) attributes standalone_in_db and serialized (especially serialized) ?

Tanguylo: After some thoughts, i guess you compute it only once. But why inheriting from Schemas when you already have a class Class that inherits from Schemas and you could use as parent ?

Tanguylo (approved):

dessia_common/schemas/core.py
GhislainJ: I found this syntax while trying to find a way to prevent pycharm from complaining about return types not validating type hints. I wanted to define that the return type should be an extension of the base chunk, for ex :+1: 

```py
base_chunk = {a: 0, b: 1}
chunk = {**base_chunk, c: 2}
```

But pycharm or mypy doesn't seem to understand that unpacking a dict into another can set the right keys in the result dict which would then validate properly the schema. That's where i should have written : 

`chunk = cast(ExtendedType, {**base_chunk, c: 2})`

Which is the equivalent to typescript's "as", but i am reluctant to do so.

I might reuse this syntax in other schemas

PR #566, opened 2023-03-23
Merge dev_docx_generator into dev

younesdessia:
- [ ] Bug fix
- [X] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:
Add new module to write docx file

### DIFF
--- a/scripts/ci_scripts.py
+++ b/scripts/ci_scripts.py
@@ -18,6 +18,7 @@ scripts = [
     "markdowns.py",
     "checks.py",
     "unit_tests.py",
+    "docx_writer.py",
 
     # Workflows
     "workflow/blocks.py",


dessia_common/core.py
```diff
@@ -874,7 +874,7 @@ def get_booleans_index(self, values: List[DessiaObject]):
         :param values: List of DessiaObjects to filter
         :type values: List[DessiaObject]
 
-        :return: `list of length `len(values)` where elements are `True` if kept by the filter, otherwise `False`.
+        :return: `list of length `len(values)` where the elements are `True` if kept by the filter, otherwise `False`.
```
GhislainJ, line 877: I don't think docstring should be changed here; Old version seem more "english" to me

dessia_common/document_generator.py
```diff
@@ -0,0 +1,93 @@
...
+    """ write a docx file. """
+
+    def __init__(self, filename: str):
+        self.filename = filename
+        self.doc = docx.Document()
+        self.headings = None
+
+    def add_headings(self, headings: List[Tuple[str, int]]) -> 'DocxWriter':
+        """ Add a list of headings to the document. """
+        new_doc = self.doc
```
GhislainJ, line 19: There is no need to add "new_" prefix over `doc` and `writer` variables here. Consider naming them `document` and `writer`

dessia_common/document_generator.py
```diff
@@ -0,0 +1,93 @@
...
+            new_doc.add_heading(heading[0], heading[1])
+        new_writer = DocxWriter(self.filename)
+        new_writer.doc = new_doc
+        new_writer.headings = headings
+        new_writer.add_page_breaks(1)
+        return new_writer
+
+    def add_paragraphs(self, paragraphs: List[str]) -> 'DocxWriter':
+        """ Add a list of paragraphs to the document. """
+        new_doc = self.doc
```
GhislainJ, line 30: Why do you need to assign this variable ? Is this in order to assign a temporary variable ? it feels weird to not just use "self.doc" in the method

dessia_common/document_generator.py
```diff
@@ -0,0 +1,93 @@
...
+        for heading in headings:
+            new_doc.add_heading(heading[0], heading[1])
+        new_writer = DocxWriter(self.filename)
+        new_writer.doc = new_doc
+        new_writer.headings = headings
+        new_writer.add_page_breaks(1)
+        return new_writer
+
+    def add_paragraphs(self, paragraphs: List[str]) -> 'DocxWriter':
```
GhislainJ, line 28: Same comment about "new_" prefix here

dessia_common/document_generator.py
```diff
@@ -0,0 +1,93 @@
...
+        """ Add a list of paragraphs to the document. """
+        new_doc = self.doc
+        new_headings = self.headings
+        for paragraph in paragraphs:
+            if new_headings:
+                new_doc.add_heading(new_headings[0][0], new_headings[0][1])
```
GhislainJ, line 34: What is `heading[i][j]` ? It looks like it is an arbitrary structure. Should you create a class for heading ?

dessia_common/document_generator.py
```diff
@@ -0,0 +1,93 @@
...
+                new_headings = new_headings[1:]
+            new_doc.add_paragraph(paragraph)
+        new_writer = DocxWriter(self.filename)
+        new_writer.doc = new_doc
+        new_writer.headings = new_headings
+        return new_writer
+
+    def add_page_breaks(self, num_page_breaks: int):
+        """ Add an empty page to the document. """
```
GhislainJ, line 43: Docstring and features seem off. To me it looks like we are adding a certain number of page breaks, not just one

dessia_common/document_generator.py
```diff
@@ -0,0 +1,93 @@
...
+        for _ in range(num_page_breaks):
+            self.doc.add_page_break()
+
+    def add_table(self, rows: List[List[str]]) -> 'DocxWriter':
```
GhislainJ, line 47: Same comment about new_ prefix

dessia_common/document_generator.py
```diff
@@ -0,0 +1,93 @@
...
+            header = section.header
+            for paragraph in header.paragraphs:
+                paragraph.text = text
+                paragraph.alignment = getattr(docx.enum.text.WD_ALIGN_PARAGRAPH, align.upper())
+        else:
+            footer = section.footer
+            for paragraph in footer.paragraphs:
+                paragraph.text = text
+                paragraph.alignment = getattr(docx.enum.text.WD_ALIGN_PARAGRAPH, align.upper())
+        new_writer = DocxWriter(self.filename)
```
GhislainJ, line 76: Why do we create a new DocxWriter in every method ?

dessia_common/document_generator.py
```diff
@@ -0,0 +1,93 @@
...
+        for row in rows[1:]:
+            row_cells = table.add_row().cells
+            for i, cell in enumerate(row_cells):
+                cell.text = row[i]
+        new_writer = DocxWriter(self.filename)
+        new_writer.doc = new_doc
+        new_writer.headings = self.headings
+        return new_writer
+
+    def add_header_footer(self, text: str, is_header: bool = True, align: str = 'center') -> 'DocxWriter':
```
GhislainJ, line 62: this should be split into two methods `add_header` and `add_footer`

dessia_common/document_generator.py
```diff
@@ -0,0 +1,93 @@
...
+        """ Add a header or footer to the document. """
+        new_doc = self.doc
+        section = new_doc.sections[-1]
+        if is_header:
+            header = section.header
+            for paragraph in header.paragraphs:
```
GhislainJ, line 68: This part can be mutualized for both header and footer to avoid maintenance "cumbersomeness"

dessia_common/document_generator.py
```diff
@@ -0,0 +1,93 @@
...
+            for paragraph in footer.paragraphs:
+                paragraph.text = text
+                paragraph.alignment = getattr(docx.enum.text.WD_ALIGN_PARAGRAPH, align.upper())
+        new_writer = DocxWriter(self.filename)
+        new_writer.doc = new_doc
+        new_writer.headings = self.headings
+        return new_writer
+
+    def add_picture(self, image_path: str, width: int = None, height: int = None) \
+            -> 'DocxWriter':
```
GhislainJ, line 82: Is this linebreak necessary ? It doesn't look like the line break the 120 characters threshold

GhislainJ (changes requested):
This loo,s like it only is for "local" use. How does it behave in the platform or is it supposed to ?

dessia_common/tests.py
```diff
@@ -348,7 +348,7 @@ def __init__(self, p_1: float, p_2: float, p_3: float, p_4: float, p_5: float, n
 
     @property
     def test_prop(self):
-        """Factice property for some tests."""
+        """Dummy property for some tests."""
```
GhislainJ, line 351: If you get a conflict where to other where "Fake" is proposed, please keep "Dummy", as you proposed. It is way better, indeed

GhislainJ (changes requested):
(Don't mind this comment, it is just to trigger the "Request Change" again after a handling mistake from my part"

dessia_common/document_generator.py
younesdessia: it's just so as not to modify the self directly

dessia_common/document_generator.py
younesdessia: I think it's better that way !

dessia_common/document_generator.py
younesdessia: you're right, because before I had more parameters !

dessia_common/document_generator.py
younesdessia: ok !

younesdessia: I don't think it's a good idea

GhislainJ: Well, actually, the best practice here, would be to have a class for "section";

Why do you think it is a bad idea ?

dessia_common/document_generator.py
GhislainJ: For which reason ?

GhislainJ (changes requested):

GhislainJ (changes requested):

dessia_common/document_generator.py
younesdessia: because header and  footer it's the same thing just the position (the location) on the document is different

dessia_common/document_generator.py
younesdessia: I agree, we have to create another class

dessia_common/document_generator.py
GhislainJ: Yes, we agree on that point, that why we need to think about the following structure for DocxWriter : 
```

{
    document: Document,
    filename: str,
    headings: ...,
    header: Section (Or Header inheriting from Section, section could be wider),
    footer: Section
}
```

We could also have a "body" attribute, for example, and some other for other stuff.

header and footer would then share the same method that would be called from parent object to generate the right structure (DocxWriter)

Will it be a future dev for you and should I merge this PR now ?

younesdessia: Yes, I'm setting this up on my repository,
Yes, you merge it !
tell me if you want me to add the script that I sent you in PM ?

GhislainJ (approved):

PR #571, opened 2023-03-30
Save workflowrun as python script

younesdessia:
- [ ] Bug fix
- [X] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [X] No

* **Other information**:

Export wf_run as python script with input


younesdessia:
@GhislainJ 
I'm open to discussing the possibility of adding the ability to record streams to temporary files....
 Thank you.

GhislainJ:
In order to get this PR going again, we need to set in the script that user should reference their files themselves : 

```py
file1 = BinaryFile.from_file("Insert your path here")
input_values = {0: file1}
```

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
         warnings.warn("method_jsonschema method is deprecated. Use method_schema instead", DeprecationWarning)
         return self.method_schemas
 
+    @staticmethod
+    def instantiate_objects(objects):
+        """ Instantiates objects from a list of objects. """
+        class_name = f"{objects[0].__class__.__name__}"
```
GhislainJ, line 1949: Consider this : 
```py
signature = "('Set your arguments here')"
    if len(objects) > 1:
        instances = [f"\n\t{o.__class__.__name__}{signature}" for o in objects]
        return f"[{','.join(instances)}\n]"
    return f"{objects[0].__class__.__name__}{signature}"
```

Outputs : 

```py
[
	A('Set your arguments here'),
	B('Set your arguments here'),
	A('Set your arguments here')
]
```

```py
A('Set your arguments here')
```

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
         warnings.warn("method_jsonschema method is deprecated. Use method_schema instead", DeprecationWarning)
         return self.method_schemas
 
+    @staticmethod
+    def instantiate_objects(objects):
```
GhislainJ, line 1947: Why does need to be a workflow method ?
In addition to that, method name seems off

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+            for _ in objects:
+                inst = f"{class_name}('Put here your object')"
+                instances += inst + ", "
+            instances += ']'
+            return instances
+
+        return f"{class_name}('Put here your object')"
+
+    @staticmethod
+    def get_files(type_files):
```
GhislainJ, line 1961: Consider applying the same kind of function than for objects.
In addition to that, method name seems off

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+            return instances
+        return f"{file_class}.from_file('Insert your path here')"
+
+    @staticmethod
+    def is_object_sequence(seq):
+        """ Checks if a sequence contains serializable objects. """
+        return any(isinstance(element, SerializableObject) for element in seq)
+
+    @staticmethod
+    def is_file_sequence(sequence):
```
GhislainJ, line 1978: Consider using `is_file_or_file_sequence` or `is_dessia_file` from `types.py` module

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
         warnings.warn("method_jsonschema method is deprecated. Use method_schema instead", DeprecationWarning)
         return self.method_schemas
 
+    @staticmethod
+    def instantiate_objects(objects):
+        """ Instantiates objects from a list of objects. """
```
GhislainJ, line 1948: Unclear docstring

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+                inst = f"{class_name}('Put here your object')"
+                instances += inst + ", "
+            instances += ']'
+            return instances
+
+        return f"{class_name}('Put here your object')"
+
+    @staticmethod
+    def get_files(type_files):
+        """ Generates file instances from a list of file types. """
```
GhislainJ, line 1962: Unclear docstring

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+    @staticmethod
+    def is_file_sequence(sequence):
+        """ Checks if a sequence contains binary files. """
+        return any(isinstance(element, BinaryFile) for element in sequence)
+
+    def get_sequence(self, sequence):
+        """ Processes a sequence based on its content. """
+        if self.is_object_sequence(sequence):
+            value = self.instantiate_objects(sequence)
+        elif self.is_file_sequence(sequence):
```
GhislainJ, line 1986: unnecessary `elif` and `else` statements

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+        if len(type_files) > 1:
+            instances = "["
+            for _ in range(len(type_files)):
+                instances += f"{file_class}.from_file('Insert your path here')" + ", "
+            instances += ']'
+            return instances
+        return f"{file_class}.from_file('Insert your path here')"
+
+    @staticmethod
+    def is_object_sequence(seq):
```
GhislainJ, line 1973: There seems to be no need for a function that does just that.
In addition to that, please avoid variable abbreviation

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+        """ Checks if a sequence contains serializable objects. """
+        return any(isinstance(element, SerializableObject) for element in seq)
```
GhislainJ, line 1975: I have a feeling that this should be an "all" statement instead.
What happens in other cases ? (Not all or not any)

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+
+    @staticmethod
+    def is_file_sequence(sequence):
+        """ Checks if a sequence contains binary files. """
+        return any(isinstance(element, BinaryFile) for element in sequence)
+
+    def get_sequence(self, sequence):
```
GhislainJ, line 1982: Unclear name of method

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+        elif self.is_file_sequence(sequence):
+            value = self.get_files(sequence)
+        else:
+            value = repr(sequence)
+        return value
+
+    def _compute_input_values(self):
+        """ Computes the input values for blocks and non-block variables. """
+        values_nonblock_variables = []
+        values_blocks = []
```
GhislainJ, line 1995: Consider : 

```py
nbv_values = [self.input_values[i] for i, v in enumerate(self.workflow.inputs) if v in self.workflow.nonblock_variables]
block_values = [self.input_values[i] for i, v in enumerate(self.workflow.inputs) if v not in self.workflow.nonblock_variables
        return values_nonblock_variables, values_blocks

```

OR

```py
nbv_values = []
block_values = []
for i, variable in enumerate(self.workflow.inputs):
    sequence = nbv_values if variable in self.workflow.non_block_variables else block_values
    sequence.push(self.input_values[i]
return nbv_values, block_values
```

Note :
- nbv is considered ok as part of workflow's jargon
- abbreviating variable to single letter names is considered ok in list comprehension

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+        """ Processes a sequence based on its content. """
+        if self.is_object_sequence(sequence):
+            value = self.instantiate_objects(sequence)
+        elif self.is_file_sequence(sequence):
+            value = self.get_files(sequence)
+        else:
+            value = repr(sequence)
+        return value
+
+    def _compute_input_values(self):
```
GhislainJ, line 1992: Unclear method name. Maybe `_split_input_values` ? "Compute" is too large and vague

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+        """ Computes the input values for blocks and non-block variables. """
```
GhislainJ, line 1993: Unclear docstring

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+
+        # NBVs
+        for k, nbv in enumerate(self.workflow.nonblock_variables):
+            input_info["input_str"] += f"    workflow.input_index(variable_{k}): value_{k}_,\n"
+            input_info["default_value"] += self._generate_default_value(values_nonblock_variables[k], k)
+            input_info["add_import"].extend(get_schema(annotation=nbv.type_,
+                                                       attribute=SchemaAttribute(nbv.name)).get_import_names(
+                import_names=[]))
+
+        return input_info["input_str"], input_info["default_value"], input_info["add_import"]
```
GhislainJ, line 2037: Why are we not returning full dict ? There seems to be no need to turn it into a tuple OR there is no need to define a dict in the first place

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+
+        return workflow_script
+
+    def _generate_default_value(self, value, block_index, input_index=None):
+        """ Generate a value for a given input. """
+        if not input_index and input_index != 0:
+            input_index = ''
+        if isinstance(value, SerializableObject):
+            return f"\nvalue_{block_index}_{input_index} = {self.instantiate_objects([value])}"
+        if isinstance(value, (BinaryFile, StringFile)):
```
GhislainJ, line 2056: Consider using is_dessia_file function

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+
+    def _to_script(self, workflow_script):
+        """ Computes elements for a to_script interpretation. """
+        input_info = {
+            "input_str": "",
+            "default_value": "",
+            "add_import": [],
+        }
+
+        block_input = "block_{}.inputs[{}]"
```
GhislainJ, line 2012: As it is only use once, you could directly set this as an f-string in the loop statement (instead of formatting)

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+
+        workflow_script.declaration = f"{workflow_script.declaration}\n{default_value}\n\ninput_values =" \
+                                      f" {{\n{input_str}}}\nworkflow_run = workflow.run(input_values=input_values)\n"
+        workflow_script.imports.extend(add_import)
+
+        return workflow_script
+
+    def _generate_default_value(self, value, block_index, input_index=None):
+        """ Generate a value for a given input. """
+        if not input_index and input_index != 0:
```
GhislainJ, line 2052: What is the usecase of this ? Why not just `if input_index is None` ?

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+        # Blocks
+        index_ = 0
+        for j, block in enumerate(self.workflow.blocks):
+            for i, input_ in enumerate(block.inputs):
+                input_key = block_input.format(j, i)
+                if input_key not in workflow_script.declaration:
+                    input_info["input_str"] += f"    workflow.input_index({input_key}): value_{j}_{i},\n"
+                    input_info["default_value"] += self._generate_default_value(values_blocks[index_], j, i)
+                    index_ += 1
+                    input_info["add_import"].extend(get_schema(annotation=input_.type_,
```
GhislainJ, line 2025: Please simplify this line by assigning variables to schema and eventually imports

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,138 @@ def _method_jsonschemas(self):
...
+                                                               attribute=SchemaAttribute(input_.name)).get_import_names(
+                        import_names=[]))
+
+        # NBVs
+        for k, nbv in enumerate(self.workflow.nonblock_variables):
+            input_info["input_str"] += f"    workflow.input_index(variable_{k}): value_{k}_,\n"
+            input_info["default_value"] += self._generate_default_value(values_nonblock_variables[k], k)
+            input_info["add_import"].extend(get_schema(annotation=nbv.type_,
```
GhislainJ, line 2033: Please simplify this line by assigning variables to schema and eventually imports

GhislainJ (changes requested):

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: because if input_index is zero it is considered None, and if it's equal to zero I need it

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: I did this (dict) because I had pyling error, I'll update it !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: I did as you asked, but I had a CI problem, that's why I encapsulated everything !
I will update  !

dessia_common/workflow/core.py
younesdessia: same !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: in this case we consider that the sequence only contains SerializableObjects, I don't see why in the sequence we will have a SerializableObject and a file for example!

younesdessia:
@GhislainJ 
ready for review !

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,119 @@ def _method_jsonschemas(self):
         warnings.warn("method_jsonschema method is deprecated. Use method_schema instead", DeprecationWarning)
         return self.method_schemas
 
+    @staticmethod
+    def generate_object_input_string(objects):
```
GhislainJ, line 1947: These two methods (`generate_object_input_string` & `generate_file_input_string`) could still be functions detached from Workflow Class.

Now that you refactored them, they are really alike, and you could pass signature as an argument and simplify your calls in `process_sequence` methods

dessia_common/workflow/core.py
GhislainJ: That is why I think it is better to use an "all" statement. If you know that your sequence should be homogeneous, then you should also be conservative in your check and check for all instead of any

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,119 @@ def _method_jsonschemas(self):
...
+            instances = [f"\n\t{o.__class__.__name__}{signature}" for o in type_files]
+            return f"[{','.join(instances)}\n]"
+        return f"{type_files[0].__class__.__name__}{signature}"
+
+    @staticmethod
+    def is_object_sequence(sequence):
+        """ Checks if a sequence contains serializable objects. """
+        return any(isinstance(element, SerializableObject) for element in sequence)
+
+    def process_sequence(self, sequence):
```
GhislainJ, line 1969: This could also be a function, if the others are set as functions.

Consider setting them in workflow/utils.py

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,119 @@ def _method_jsonschemas(self):
...
+    def _split_input_values(self):
+        """ Split the input values for blocks and non-block variables. """
+        nbv_values = [self.input_values[i] for i, v in enumerate(self.workflow.inputs) if
+                      v in self.workflow.nonblock_variables]
+        block_values = [self.input_values[i] for i, v in enumerate(self.workflow.inputs) if
+                        v not in self.workflow.nonblock_variables]
+        return nbv_values, block_values
+
+    @staticmethod
+    def _update_imports(input_type, input_name):
```
GhislainJ, line 1986: Could be a function

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,119 @@ def _method_jsonschemas(self):
...
+        """ Update imports based on the schema of the input. """
+        schema = get_schema(annotation=input_type, attribute=SchemaAttribute(input_name))
+        imports = schema.get_import_names(import_names=[])
+        return imports
```
GhislainJ, line 1990: You could directly `return schema.get_import_names(import_names=[])`

GhislainJ (changes requested):

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: done !

dessia_common/workflow/utils.py
```diff
@@ -65,3 +69,41 @@ def nonblock_variables_to_script(nonblock_variables, prefix, imports, imports_as
...
+    """ Checks if a sequence contains serializable objects. """
+    return all(isinstance(element, SerializableObject) for element in sequence)
+
+
+def process_value(value):
+    """ Processes a value based on its content. """
+    if not value:
+        return repr(value)
+
+    sequence_length = len(value)
```
GhislainJ, line 98: What if value has no len ?

dessia_common/workflow/utils.py
```diff
@@ -65,3 +69,41 @@ def nonblock_variables_to_script(nonblock_variables, prefix, imports, imports_as
...
     return nbvs_str
+
+
+def update_imports(input_type, input_name):
+    """ Update imports based on the schema of the input. """
+    schema = get_schema(annotation=input_type, attribute=SchemaAttribute(input_name))
+    return schema.get_import_names(import_names=[])
+
+
+def generate_input_string(signature: str, sequence_length: int = 1):
```
GhislainJ, line 80: You should still pass the sequence instead of the sequence length, as you did before.

The argument you should pass is the `('Set your arguments here')` or `.from_file('Set your filepath here')` part

dessia_common/workflow/core.py
```diff
@@ -1941,6 +1943,78 @@ def _method_jsonschemas(self):
...
+        input_str, default_value, add_import = self._to_script(workflow_script)
+
+        workflow_script.declaration = f"{workflow_script.declaration}\n{default_value}\n\ninput_values =" \
+                                      f" {{\n{input_str}}}\nworkflow_run = workflow.run(input_values=input_values)\n"
+        workflow_script.imports.extend(add_import)
+
+        return workflow_script
+
+    @staticmethod
+    def _generate_default_value(value, block_index: int, input_index: int = None):
```
GhislainJ, line 1993: Could this be a function, as well ? Instead of a static method

In addition to that, consider 
```py
input_index_str = "" if input_index is None else f"_{input_index}"
```

GhislainJ (changes requested):

dessia_common/workflow/core.py
younesdessia: yes,
done !

dessia_common/workflow/utils.py
younesdessia: if value has no len, it's the first condition (same if value is None or [ ])

```py
    if not value:
        return repr(value)
```

dessia_common/workflow/utils.py
younesdessia: done !

dessia_common/workflow/utils.py
GhislainJ: This condition is not safe enough. Objects could be considered True and have no len. here is an example : 

```py
class A:
    def __bool__(self):
        return True


a = A()


if not a:
    print("False")
else:
    print("Failing")
    print(len(a))
```

dessia_common/workflow/utils.py
```diff
@@ -65,3 +69,44 @@ def nonblock_variables_to_script(nonblock_variables, prefix, imports, imports_as
...
+        return repr(value)
+
+    if is_object_sequence(value):
+        signature = "('Set your arguments here')"
+    elif is_file_or_file_sequence(value):
+        signature = ".from_file('Set your filepath here')"
+    else:
+        return repr(value)
+
+    return generate_input_string(signature, value)
```
GhislainJ, line 105: This is still not safe for non sequence objects as you evaluate len in `generate_input_string`. Consider : 

```py
    if not value: # Under certain condition (false objects ruled out from is_sequence functions) you could remove this
        return repr(value)

    if is_object_sequence(value):
        return generate_input_string("('Set your arguments here')", value)
    elif is_file_or_file_sequence(value):
        return generate_input_string(".from_file('Set your filepath here')", value)
    return repr(value)
```

If your is_object_sequences and is_file_or_file_sequence functions rule false objects out, you can even remove the first check.

dessia_common/workflow/utils.py
GhislainJ: Anyway, the fact that the statement has been removed **kinda** make all this irrelevant

GhislainJ (changes requested):

younesdessia:
@GhislainJ 
it's a mistake on my part, I thought this `value` was a list like I did at the beginning,
I fixed all that and consolidated the unit test.

GhislainJ (approved):

PR #595, opened 2023-07-24
feat: workflow memory monitoring

masfaraud:
- [ ] Bug fix
- [x] new features: monitoring memory on WorkflowRun/State
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [X] No

* **Other information**:
* Creating a ExecutionInfo class to gather variables
* removing jointjs plot for workflow


setup.py
```diff
@@ -138,23 +138,25 @@ def get_branch():
...
     ],
     install_requires=[
-        "orjson>=3.8.0",
-        "typeguard",
+        "cma",
+        "dectree",
+        "docx",
+        "humanize",
+        "matplotlib",
+        "mypy_extensions",
```
GhislainJ, line 146: Maybe we could remove mypy_extensions in the future, as we state that python version is <3.9. It was necessary for python <3.8

GhislainJ (commented):
As you stated, I think it would actually be smarter to use lists instead of dicts

PR #602, opened 2023-09-18
perf: add documentation to workflow when export as .py

younesdessia:
- [ ] Bug fix
- [ ] new features
- [X] performance 
- [ ] docs update

- [ ] Yes:
- [X] No

* **Other information**:


dessia_common/workflow/core.py
```diff
@@ -1334,7 +1334,9 @@ def to_script(self) -> str:
         script_imports = self_script.imports_to_str()
 
         return f"{script_imports}\n" \
-               f"{self_script.declaration}"
+               f"{self_script.declaration}\n" \
+               f'documentation = """{self.documentation}"""' \
```
GhislainJ, line 1338: I think we should avoid using triple quote inside a plain script.
Consider `f"documentation = '{self.documentation}'"` instead

GhislainJ (changes requested):

dessia_common/workflow/core.py
younesdessia: image

with using `f"documentation = '{self.documentation}'"` :
image


image

GhislainJ (approved):

PR #624, opened 2023-11-02
Feat/handle ivv

GhislainJ:
- [ ] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:


dessia_common/workflow/blocks.py
```diff
@@ -84,11 +85,11 @@ class InstantiateModel(Block):
...
     """
 
-    def __init__(self, model_class: Type, name: str = "", position: Tuple[float, float] = None):
+    def __init__(self, model_class: Type, name: str = "Instantiate Model", position:  Position = (0, 0)):
         self.model_class = model_class
         inputs = []
         inputs = set_inputs_from_function(self.model_class.__init__, inputs)
-        outputs = [TypedVariable(type_=self.model_class, name="Model")]
+        outputs = [Variable(type_=self.model_class, name="Model")]
         Block.__init__(self, inputs, outputs, name=name, position=position)
```
Thiefaine-dev, line 93: Same comment here

dessia_common/workflow/blocks.py
```diff
@@ -164,6 +173,18 @@ def equivalent(self, other: 'ClassMethod'):
         same_method = self.method_type.name == other.method_type.name
         return Block.equivalent(self, other) and same_class and same_method
```
Thiefaine-dev, line 174: Same comment here

dessia_common/workflow/blocks.py
```diff
@@ -1282,14 +1392,14 @@ class Archive(Block):
...
+    def __init__(self, number_exports: int = 1, filename: str = "archive", name: str = "Archive",
+                 position:  Position = (0, 0)):
         self.number_exports = number_exports
         self.filename = filename
         self.extension = "zip"
         self.text = False
         inputs = [Variable(name=f"Export {i}") for i in range(number_exports)]
-        inputs.append(TypedVariableWithDefaultValue(type_=str, default_value=filename, name="Filename"))
+        inputs.append(Variable(type_=str, default_value=filename, name="Filename"))
         Block.__init__(self, inputs=inputs, outputs=[Variable(name="Archive")], name=name, position=position)
```
Thiefaine-dev, line 1403: A lot of init use `NameOfClass.__init__()`. Does it have a reason? If not it's preferable to use `super().__init__`

Thiefaine-dev (changes requested):

dessia_common/workflow/blocks.py
GhislainJ: You're right ! For a lot of cases, it is old code legacy, but for the new ones, that is just me being careless. I tend to use `super().__init__()` more often, but sometimes I just slip into `ClassName.__init__()`.

I'll perform some changes

PR #660, opened 2024-01-17
Fix/ref path cycles

GhislainJ:
- [ ] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:


dessia_common/core.py
```diff
@@ -432,7 +432,7 @@ def _display_from_selector(self, selector: str) -> DisplayObject:
         track = ""
         try:
             data = attrmethod_getter(self, display_setting.method)(**display_setting.arguments)
-        except:
+        except Exception:
```
younesdessia, line 435: Can't we put the real exception here?

younesdessia (approved):

dessia_common/core.py
GhislainJ: Actually, this is the wanted behavior in that case.
Whatever happens here, we don't want the platform to completely crash if a display cannot be computed.
As we read user code, we don't know exactly what kind of error can happen. This is why there is 9 occurrences allowed of `broad-exception-caught` it `code_pylint.py`. I set Exception to remove the `bare-except` entry which is essentially the same. This just feels a bit less "wrong"

PR #661, opened 2024-01-18
Fix/Update: export workflow to script

younesdessia:
- [X] Bug fix
- [ ] new features
- [X] performance 
- [ ] docs update

- [ ] Yes:
- [X] No

* **Other information**:


dessia_common/workflow/blocks.py
```diff
@@ -1452,3 +1472,11 @@ def get_attribute_type(attribute_name: str, parameters):
     if parameter.annotation == inspect.Parameter.empty:
         return None 
     return parameter.annotation
+
+
+def to_script_selector(display) -> ToScriptElement:
```
GhislainJ, line 1477: Should be a Display method

dessia_common/workflow/core.py
```diff
@@ -106,6 +107,65 @@ def copy(self, deep: bool = False, memo=None):
         else:
             copied_default_value = UNDEFINED
         return Variable(type_=self.type_, default_value=copied_default_value, name=self.name)
+    
+    def process_heterogeneous_sequence(self, schema, import_str_list):
```
GhislainJ, line 111: You do not use self here, but instead a  lot of schema, while having a lot of "isinstance" stuff, so I guess this should be a method of schema that gives import depending on the type of schema.

This goes for all your `process_*` methods here

tests/test_workflow/test_to_script.py
```diff
@@ -1,100 +1,125 @@
...
+
+from dessia_common.typings import MethodType, MarkdownType, PlotDataType
+from dessia_common.workflow.core import Pipe, Workflow, Variable
+from dessia_common.workflow.blocks import InstantiateModel, ModelMethod, PlotData, Markdown, Export
+from dessia_common.decorators import markdown_view, plot_data_view
+from dessia_common.files import BinaryFile
+from dessia_common.core import DessiaObject
+
+
+class MyObj(DessiaObject):
```
GhislainJ, line 12: Avoid abbreviation

tests/test_workflow/test_to_script.py
```diff
@@ -1,100 +1,125 @@
...
+    def __init__(self, name: str = "", number_connection: int = 2, file: BinaryFile = None):
+        self.number_connection = number_connection
+        self.file = file
+        super().__init__(name=name)
+
+
+class MyExp(DessiaObject):
```
GhislainJ, line 19: Avoid abbreviation

tests/test_workflow/test_to_script.py
```diff
@@ -1,100 +1,125 @@
...
+                        pipe_1, pipe_2, pipe_3, pipe_4, pipe_5, pipe_6, pipe_7,
+                        pipe_8, pipe_9, pipe_10, pipe_11
+                    ],
+                    output=blocks[1][0].outputs[0], name="TestExport")
+
+
+class WorkflowToScriptTest(unittest.TestCase):
+
+    def test_blocks(self):
+        for block in blocks:
```
GhislainJ, line 112: You could use parameterized. Please check other test files that use it

GhislainJ (changes requested):

dessia_common/workflow/blocks.py
```diff
@@ -749,10 +749,30 @@ def evaluate(self, values, **kwargs):
...
 
+    @property
+    def display_class_name(self):
+        """ Get the class name of the current instance. """
+        return full_classname(self).split('.')[-1]
+
+    @property
+    def selector_class_name(self):
+        """ Get the class name of the Selector. """
+        return full_classname(self.selector).split('.')[-1]
```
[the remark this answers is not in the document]
younesdessia: done !

dessia_common/workflow/blocks.py
younesdessia: done !

dessia_common/workflow/core.py
younesdessia: refactored !

done

tests/test_workflow/test_to_script.py
younesdessia: done !

tests/test_workflow/test_to_script.py
younesdessia: done

tests/test_workflow/test_to_script.py
younesdessia: done

dessia_common/schemas/core.py
```diff
@@ -940,6 +954,17 @@ def classes_are_standalone_consistent(self) -> PassedCheck:
               f"are not consistent. They should be all standalone in db or none of them should."
         return WrongType(msg)
 
+    def get_import_names(self, import_names):
+        """ Process UnionProperty and get/update the import names list."""
+        import_name = f"{self.origin.__module__}.{self.origin._name}"
```
GhislainJ, line 959: As we discussed yesterday, this could be set as a cached_property in TypingProperty class

dessia_common/schemas/core.py
```diff
@@ -1042,6 +1067,15 @@ def ellipsis_has_exactly_two_args(self) -> PassedCheck:
             return WrongNumberOfArguments(msg)
         return PassedCheck(f"{self.check_prefix}is not an ill-defined ellipsed tuple : '{self.annotation}'.")
 
+    def get_import_names(self, import_names):
```
GhislainJ, line 1070: This is the same for all classes inheriting from TypingProperty and should probably be mutualized there (caution, Union is a bit different, but I think it could be the same. This has to be checked)

GhislainJ (changes requested):

dessia_common/schemas/core.py
younesdessia: We can't set this as a cached_property because I need to give import_names list

dessia_common/schemas/core.py
younesdessia: done !

GhislainJ:
As another element, I think we could add the "label" attribute of Variable to its to_script, so that someone who previously set labels in the workflow builder doesn't lose them by going to script and generating a new platform workflow

dessia_common/schemas/core.py
younesdessia: it's done as Property

GhislainJ (approved):

PR #675, opened 2024-02-23
Fix: duplicate result name nbv

younesdessia:
- [X] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [X] No

* **Other information**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - ModelMethod : Call the right dict_to_object of MethodType
 - MultiPlot : Add selector_name MultiPlot script
+- Result name : Update RESULT_VARIABLE_NAME
 - WorkflowRun : Add markdown to display_settings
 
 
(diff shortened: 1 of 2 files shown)


younesdessia:
@GhislainJ 
as we have the possibility of putting a label on the Varaiable, I think we can do that

younesdessia:
Yes but if we use _result_name_ as a name, the _find_name function adds a second nbv (Result Name)

I think that in this case, you have to put the same name, like that we will detect it and if the user wants to change he changes the label!

```py
    def _find_name(self):
        found_name = False
        i = 0
        all_nbvs = self.nonblock_variables + self.detached_variables
        while not found_name and i < len(all_nbvs):
            variable = all_nbvs[i]
            found_name = variable.name == RESULT_VARIABLE_NAME
            i += 1
        if not found_name:
            self.detached_variables.insert(0, NAME_VARIABLE)
```

younesdessia:
I'm closing this PR because I understood why I had this problem,

in fact I defined a nbv locally with the name `Result Name` that's why dessia_common adds a new one to me whereas when we let the front end do it it works well.

PR #677, opened 2024-02-28
Fix: object_default/schema

younesdessia:
- [X] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [X] No

* **Other information**:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## 0.16.2
+
+### Fix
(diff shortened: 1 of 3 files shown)


GhislainJ (changes requested):
We need more context about this, as None as default value should directly happen with objects. We should normally have an Optional schema

younesdessia:
Another fix, replace None value of `default_value` with `UNDEFINED` in `SchemaAttribute` class:  

image

GhislainJ (approved):

GhislainJ:
Approved but adding tests are needed to merge

GhislainJ (changes requested):
Tests using you simplified repro case should be added

younesdessia:
Done !

GhislainJ (approved):

PR #679, opened 2024-03-07
Fix: add block index to workflowrun zip

younesdessia:
- [x] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [x] No

* **Other information**:


GhislainJ (commented):
What is the issue ?

younesdessia:
Hi,
In the Workflow object, if we have export blocks, we must pass index_block to the export method

younesdessia:
@GhislainJ 

here is a simple example to reproduce this issue !


```py
from dessia_common.core import DessiaObject


class ObjectTest(DessiaObject):

    def __init__(self, name: str = "", measure: float = 1.1):
        self.measure = measure
        DessiaObject.__init__(self, name=name)


class InstantiateObject(DessiaObject):

    def __init__(self, name: str = ""):
        DessiaObject.__init__(self, name=name)

    def compute(self, measure: float = 1):
        return ObjectTest(measure=measure)

```


```py
from dessia_common.typings import MethodType
from dessia_common.workflow.blocks import InstantiateModel, ModelMethod, Export
from dessia_common.workflow.core import Pipe, Workflow
from dessia_common.export_test import InstantiateObject

block_0 = InstantiateModel(model_class=InstantiateObject, name="InstantateObject")
block_1 = ModelMethod(method_type=MethodType(InstantiateObject, 'compute'), name="compute")
txt_method = MethodType(class_=InstantiateObject, name="save_to_stream")
export_txt = Export(method_type=txt_method, text=True, filename="export_json", extension="json", name="Export JSON")

blocks = [block_0, block_1, export_txt]


pipe_1 = Pipe(block_0.outputs[0], block_1.inputs[0])
pipe_2 = Pipe(block_0.outputs[0], export_txt.inputs[0])

workflow = Workflow(blocks, [pipe_1, pipe_2], output=block_1.outputs[0], name="test")

input_values = {

}

workflow_export_run = workflow.run(input_values)
workflow_export_run.zip_settings() # error block index when evaluate txt_method export method (block_index!)



```

dessia_common/workflow/core.py
```diff
@@ -1965,6 +1965,20 @@ def to_markdown(self, **kwargs) -> str:
         return template.substitute(name=self.name, workflow_name=self.workflow.name,
                                    output_table=output_table, execution_info=execution_info)
 
+    def zip_settings(self):
+        """ Returns a list of streams that contain different exports of the objects. """
+        streams = []
+        for export_format in self._export_formats():
```
GhislainJ, line 1971: Should only loop on block export formats

dessia_common/workflow/core.py
```diff
@@ -1965,6 +1965,20 @@ def to_markdown(self, **kwargs) -> str:
...
+            if export_format.extension != "zip":
+                method_name = export_format.method_name
+                stream_class = StringFile if export_format.text else BinaryFile
+                stream = stream_class(filename=f"export.{export_format.extension}")
+                block_index = export_format.args.get('block_index', None)
+                _ = getattr(self, method_name)(stream, block_index) if block_index\
```
GhislainJ, line 1977: If only block export formats are in use, the negative part of this test is not needed as every export format should have an index

GhislainJ (changes requested):
Requested changes following our discussion yesterday

dessia_common/workflow/core.py
younesdessia: Done !

PR #693, opened 2024-05-15
Schemas for Literals

GhislainJ:
- [ ] Bug fix
- [x] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:


dessia_common/core.py
```diff
@@ -603,33 +603,12 @@ def to_zip(self, filepath: str):
         with open(filepath, 'wb') as file:
             file.write(archive.getbuffer())
 
-    @classmethod
```
younesdessia, line 606: how to recover these dev ?

younesdessia (approved):

dessia_common/core.py
GhislainJ: @younesdessia I reverted the problematic changes. I don't know why, but I previously decided to use this PR as a way to bring back master fixes to dev, which is a terrible decision. In the meantime I had an hotfix branch that I wrongly merged with dev, containing these changes, that I reverted (so that dev feature wouldn't reach master within a patch).
This reverted your devs.

But anyway, again, that was a terrible decision. I think the current state of the branch is OK (including patches brought back to dev), if you want to review again

Thiefaine-dev:
LGTM

PR #715, opened 2024-11-06
Feat/workflow wizard

GhislainJ:
- [ ] Bug fix
- [ ] new features
- [ ] performance 
- [ ] docs update

- [ ] Yes:
- [ ] No

* **Other information**:

### DIFF
--- a/dessia_common/core.py
+++ b/dessia_common/core.py
@@ -91,16 +91,8 @@ class DessiaObject(SerializableObject):
     _init_variables = None
     _allowed_methods = []
 
-    def __init__(self, name: str = '', **kwargs):
+    def __init__(self, name: str = ''):
         self.name = name
-        if kwargs:
-            warnings.warn(('Providing attributes to DessiaObject __init__ to be stored in self is deprecated\n'
-                           + 'Please store your attributes by yourself in your init'),
-                          DeprecationWarning)
-
-        # The code below has shown to be inefficient and should be remove in future version (0.16?)
-        for property_name, property_value in kwargs.items():
-            setattr(self, property_name, property_value)
 
     def base_dict(self):
         """ Base dict of the object, with just its name. """
--- a/dessia_common/core.py
+++ b/dessia_common/core.py
@@ -199,12 +199,8 @@ class DessiaObject(SerializableObject):
         warnings.warn("method_jsonschema method is deprecated. Use method_schema instead", DeprecationWarning)
         return self.method_schemas
 
-    def method_dict(self, method_name: str, method_jsonschema=None):
+    def method_dict(self, method_name: str):
         """ Return a default dict for the given method name. """
-        if method_jsonschema is not None:
-            warnings.warn("method_jsonschema argument is deprecated and its use will be removed in a future version."
-                          " Please remove it from your function call. Method name is sufficient to get schema",
-                          DeprecationWarning)
         schema = self.raw_method_schemas[method_name]
         return schema.default_dict()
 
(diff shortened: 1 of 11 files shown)


dessia_common/schemas/core.py
```diff
@@ -255,14 +312,24 @@ def from_type(cls, annotation: Type[T], attribute_name: str) -> 'Schema':
...
+        step = SchemaStep(annotations=annotations, attributes=[attribute], label="Step 1")
+        return cls(steps=[step])
 
     @classmethod
     def from_serialized_type(cls, type_: str, attribute_name: str) -> 'Schema':
         """ Compute a Schema directly from the serialized value of the annotation. """
         annotation = deserialize_annotation(type_)
         return cls.from_type(annotation=annotation, attribute_name=attribute_name)
 
+    def find_attribute_step(self, attribute_name: str):
```
Thiefaine-dev, line 324: Is it used?

dessia_common/workflow/core.py
```diff
@@ -46,22 +47,61 @@ class Variable(DessiaObject):
...
+            return False
+        if self.has_default_value:
+            return True
+        raise ValueError(f"Variable '{self.name}' of type '{self.type_}' and labelled '{self.label}' "
+                         f"is configured as 'Locked', but it does not implement a default value.")
+
+    @locked.setter
+    def locked(self, value: bool):
+        raise AttributeError("Attribute 'locked' of Variable cannot be changed without ensuring a default value is "
+                             "defined. Please use 'lock()' and 'unlock()' methods instead")
```
Thiefaine-dev, line 73: Why setting a setter if the only thing it does is raising an error?

dessia_common/workflow/core.py
```diff
@@ -46,22 +47,61 @@ class Variable(DessiaObject):
...
+
+    def lock(self, value: T = UNDEFINED):
+        """ Lock a variable while ensuring a default value is defined. """
+        if value is UNDEFINED and not self.has_default_value:
+            raise ValueError("Locking a Variable must implement a default value that is not UNDEFINED")
```
Thiefaine-dev, line 78: In witch case value can be UNDEFINED?

Thiefaine-dev (commented):
Nice PR. Why do not directly do the pr feat/workflow-serialization? It includes changes from this pr and don't make it as longer as it is already no?

dessia_common/workflow/core.py
GhislainJ: This is a common way in python to prevent user from directly editing the variable. 

WIthout a getter/setter, you could only make your variable private by convention by adding a `_` as a prefix, but you cannot actually prevent user from changing its value or accessing. This is the only way that I know of.

If users want to change the `locked` attribute, they have to call the `lock` method to do so

dessia_common/core.py
```diff
@@ -91,16 +91,8 @@ class DessiaObject(SerializableObject):
     _init_variables = None
     _allowed_methods = []
 
-    def __init__(self, name: str = '', **kwargs):
+    def __init__(self, name: str = ''):
```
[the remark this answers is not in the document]
GhislainJ: Two things here : 
- Yeah, you are tight about the breaking change at this line. Kwargs being removed means that people using must change their code. I might still change my mind about this change.
- About the jsonschema, that is not a breaking change technically speaking as it has been deprecated for a while now, and we use the Schemas, now

dessia_common/workflow/core.py
GhislainJ: UNDEFINED is here to describe the absence of a default value (different concept than None, as one could use None for default values, with list for example)

In your Workflow, if block inputs don't have default values defined by the code, then the default value is UNDEFINED.
Editing the workflow on the frontend allows for default value editing. In this case one can save a default value for each input, which would then replace the UNDEFINED here

dessia_common/schemas/core.py
GhislainJ: I'll need to check

